PR #4, opened 2021-12-17
feat: [MICROBA-1618] Add top navigation tabs for bulk email tool

Tj-Tracy:
image
image


justinhynes:
@Tj-Tracy Are there any screenshots we can add for visual context?

src/components/bulk-email-tool/api.js
```diff
@@ -2,8 +2,15 @@ import { getConfig } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 
 const InstructorApiBaseUrl = `${getConfig().LMS_BASE_URL}/api/instructor/v1`;
+const courseHomeBaseUrl = `${getConfig().LMS_BASE_URL}/api/course_home/course_metadata`;
```
justinhynes, line 5: Random question. I don't know much (anything) about this particular endpoint. Do you know if there is a difference between the `api/course_home/course_metadata/` endpoint and the `api/course_home/v1/course_metadata` endpoint? Do they both provide the same info? Is one marked for deprecation or anything?

src/index.jsx
```diff
@@ -2,9 +2,9 @@ import 'core-js/stable';
 import 'regenerator-runtime/runtime';
 
 import {
-  APP_INIT_ERROR, APP_READY, subscribe, initialize,
+  APP_INIT_ERROR, APP_READY, subscribe, initialize, mergeConfig,
 } from '@edx/frontend-platform';
-import { AppProvider, ErrorPage, PageRoute } from '@edx/frontend-platform/react';
+import { AppProvider, AuthenticatedPageRoute, ErrorPage } from '@edx/frontend-platform/react';
```
justinhynes, line 7: Dumb question... is there a big significance between `PageRoute` and `AuthenticatedPageRoute`? Does `AuthenticatedPageRoute` mean the page is only accessible if you're logged in/not anonymous?

src/index.jsx
```diff
@@ -34,6 +36,11 @@ subscribe(APP_INIT_ERROR, (error) => {
 });
 
 initialize({
+  config: () => {
+    mergeConfig({
+      STUDIO_BASE_URL: process.env.STUDIO_BASE_URL || null,
```
justinhynes, line 41: What do we need the Studio URL for? From a quick glance, I thought all the nav points we are adding are for places in the LMS and not the CMS?

src/components/bulk-email-tool/api.js
Tj-Tracy: AFAIK, they both provide the same data/use the same view code. I just chose the versioned endpoint because I assume that is the most up-to-date and maintained web api. Neither seem depreciated.

src/index.jsx
Tj-Tracy: Correct!

src/index.jsx
Tj-Tracy: Good point. This was needed when we had the full instructor page toolbar nav. Now that this is only the course navigation this is no longer needed.

src/components/bulk-email-tool/api.js
justinhynes: It doesn't look like we are using the versioned API above for `course_metadata`? I think that API would be the `api/course_home/v1/course_metadata`?

Tj-Tracy: oh you're right! Typo -- good catch

src/index.jsx
Tj-Tracy: We can always restrict it based on what returns from `courseHomeMetaData` api. That returns an `is_staff` boolean that we can use to restrict this page from the user.

justinhynes (approved):
LGTM! Thanks for the answers and updates.

PR #5, opened 2022-01-05
feat: Add some basic scaffolding for the bulk email tool task manager section of the comms MFE

justinhynes:
[MICROBA-1621]
* Adds a section for the pending tasks section of the task manager
* Adds a section for the bulk course email task history data
* Adds a section for the bulk course email content history data

image
image

### DIFF
--- a/src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailTaskManager.jsx
+++ b/src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailTaskManager.jsx
@@ -1,0 +1,36 @@
+import React from 'react';
+
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import BulkEmailContentHistory from './BulkEmailContentHistory';
+import BulkEmailPendingTasks from './BulkEmailPendingTasks';
+import BulkEmailTaskHistory from './BulkEmailTaskHistory';
+
+export default function BulkEmailTaskManager() {
+  return (
+    <div className="px-5">
+      <div>
+        <h2 className="h3">
+          <FormattedMessage
+            id="bulk.email.pending.tasks.section.heading"
+            defaultMessage="Pending Tasks"
+            description="A section to see pending and executing Instructor Tasks"
+          />
+        </h2>
+        <BulkEmailPendingTasks />
+      </div>
+      <div>
+        <h2 className="h3">
+          <FormattedMessage
+            id="bulk.email.task.manager.heading"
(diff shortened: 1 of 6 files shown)


justinhynes:
Screenshot of the tool from the Instructor Dashboard today for comparison:
image

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailTaskManager.jsx
```diff
@@ -0,0 +1,36 @@
...
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import BulkEmailContentHistory from './BulkEmailContentHistory';
+import BulkEmailPendingTasks from './BulkEmailPendingTasks';
+import BulkEmailTaskHistory from './BulkEmailTaskHistory';
+
+export default function BulkEmailTaskManager() {
+  return (
+    <div className="px-3">
+      <div>
+        <p className="h3">
```
Tj-Tracy, line 12: Why not just use an `h3` tag here?

Tj-Tracy (approved):
Not important or necessary for this PR, but when you get the chance pad out the section a bit from the border. The current UI:

image

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailTaskManager.jsx
justinhynes: I was trying to think about a11y concerns. I wasn't really sure how the different sections were going to get implemented so I just styled it _like_ an H3 for now. I _believe_ that headers need to go from largest to smallest, and that we can only have one h1, but I wasn't sure how the styles were going to fall since we are doing each section in chunks. 

Let me know if I am overthinking this :D.

Tj-Tracy: correct! they do go largest -> smallest, but you can skip numbers. There will eventually be an `h1` tag on this page, so setting these to `h3` I think is fine. Otherwise I believe screen readers won't scan them as headers, which I think we'd want in this case.

justinhynes: Ah, ok! I catch what you're saying now. Thanks. I'll set this to a header tag. I took a look at the original tool and I think it may be more appropriate to set this as a <h2> then but I think I will keep the h3 styling.

PR #6, opened 2022-01-05
feat: protect bulk email from non staff

Tj-Tracy:
This will protect the bulk email tool and ONLY the bulk email tool from non-staff users if they know the direct URL. this will not protect the entire app from non-staff entirely.

This PR adds a spinner, and then shows the app to the user if `is_staff` is true. Otherwise, it will show an error page.

### DIFF
--- a/src/components/bulk-email-tool/BulkEmailTool.jsx
+++ b/src/components/bulk-email-tool/BulkEmailTool.jsx
@@ -1,6 +1,8 @@
 import React, { useState, useEffect } from 'react';
 
 import { useParams } from 'react-router-dom';
+import { Spinner } from '@edx/paragon';
+import { ErrorPage } from '@edx/frontend-platform/react';
 import BulkEmailRecepient from './BulkEmailRecepient';
 import BulkEmailBody from './BulkEmailBody';
 import BulkEmailTaskManager from './bulk-email-task-manager/BulkEmailTaskManager';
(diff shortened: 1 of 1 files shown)


src/components/bulk-email-tool/BulkEmailTool.jsx
```diff
@@ -10,30 +12,49 @@ import { getCourseHomeCourseMetadata } from './api';
...
 
-  const [courseTabData, setCourseTabData] = useState([]);
+  const [courseMetadata, setCourseMetadata] = useState();
 
   useEffect(() => {
     async function fetchTabData() {
       const data = await getCourseHomeCourseMetadata(courseId);
-      const { tabs } = data;
-      setCourseTabData([...tabs]);
+      const { tabs, is_staff: isStaff } = data;
```
justinhynes, line 20: Would this cover a user who has the `instructor` role? Or `isStaff` dedicated solely to site operators/admin?

Tj-Tracy: `isStaff` in this case is given out to people who are on the "staff" list found in studio for the course run. afaik there's no "instructor" role, just admin and staff (this boolean covers both). The original permissions used by the instructor dashboard are slightly different: https://github.com/edx/edx-platform/blob/7f5b4a6a47bfc663e3f980dc6ca2d519feab6253/lms/djangoapps/instructor/permissions.py#L57

justinhynes: OK! Sounds good to me. Thanks for looking into this!

justinhynes (approved):

PR #8, opened 2022-01-07
feat: [Microba-1620] Add email form to BulkEmailTool

Tj-Tracy:
Adds the bulk email body, subject, and submit button to the bulk email tool. Not yet wired up.

image

image

### DIFF
--- a/src/components/bulk-email-tool/BulkEmailBody.jsx
+++ b/src/components/bulk-email-tool/BulkEmailBody.jsx
@@ -1,5 +1,25 @@
-import React from 'react';
+import React, { useRef } from 'react';
+import { Button, Form } from '@edx/paragon';
+
+import TextEditor from './TextEditor';
 
 export default function BulkEmailBody() {
-  return <div />;
+  const editorRef = useRef(null);
+  const onInit = (event, editor) => { editorRef.current = editor; };
+
+  return (
+    <div className="w-100 m-auto p-lg-4 p-2.5">
+      <Form>
+        <Form.Group controlId="emailSubject">
+          <Form.Label>Subject:</Form.Label>
+          <Form.Control className="w-lg-50" />
+        </Form.Group>
+        <Form.Group controlId="emailBody">
+          <Form.Label>Body:</Form.Label>
+          <TextEditor onInit={onInit} />
+        </Form.Group>
+        <Button variant="primary" type="submit">Submit</Button>
+      </Form>
+    </div>
+  );
 }
--- a/src/components/bulk-email-tool/BulkEmailBody.jsx
+++ b/src/components/bulk-email-tool/BulkEmailBody.jsx
@@ -1,6 +1,7 @@
 import React, { useRef } from 'react';
 import { Button, Form } from '@edx/paragon';
 
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import TextEditor from './TextEditor';
 
 export default function BulkEmailBody() {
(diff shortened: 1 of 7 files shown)


justinhynes:
Quick question -- the number of options in the format bar seems to be smaller (visually) compared to what is available today in the Instructor Dashboard. 

Are we able to provide feature parity between the experience today and what will be in the MFE in terms of formatting options?

src/components/bulk-email-tool/BulkEmailBody.jsx
```diff
@@ -1,5 +1,25 @@
...
+      <Form>
+        <Form.Group controlId="emailSubject">
+          <Form.Label>Subject:</Form.Label>
+          <Form.Control className="w-lg-50" />
+        </Form.Group>
+        <Form.Group controlId="emailBody">
+          <Form.Label>Body:</Form.Label>
+          <TextEditor onInit={onInit} />
+        </Form.Group>
+        <Button variant="primary" type="submit">Submit</Button>
```
justinhynes, line 21: Should we be doing the i18n work now? Otherwise we will need to come back later and re-do this, right? This goes for line 14 (`subject`) and line 18 (`Body`) as well.

src/components/bulk-email-tool/TextEditor.jsx
```diff
@@ -0,0 +1,59 @@
...
+  } = props;
+
+  return (
+    <Editor
+      initialValue=""
+      init={{
+        selector: 'textarea#editor',
+        height: 600,
+        plugins: 'advlist code emoticons link lists table',
+        toolbar: 'bold italic | bullist numlist | link emoticons',
```
justinhynes, line 33: Oh cool, this looks really extensible.

Tj-Tracy, line 33: Yeah I left it slim for now, but a ton of plugins are available -- plus we can write any we need.

src/components/bulk-email-tool/BulkEmailBody.jsx
Tj-Tracy: Good point. Working on it now 🔧

src/components/bulk-email-tool/TextEditor.jsx
justinhynes: Do you think we should do the research to compare what we have now and get parity, or do you see this as something we do later? If it is later, should we create a ticket so we don't miss it?

Tj-Tracy: I'm not sure we need to worry about this now. I'll create a subtask to check the plugins and make it a part of this ticket.

src/components/bulk-email-tool/BulkEmailBody.jsx
```diff
@@ -1,5 +1,44 @@
...
+
+  return (
+    <div className="w-100 m-auto p-lg-4 p-2.5">
+      <Form>
+        <Form.Group controlId="emailSubject">
+          <Form.Label>
+            <FormattedMessage
+              id="bulk.email.subject.label"
+              defaultMessage="Subject:"
+              description="Email subject line input label. Meant to have colon or equivilant punctuation."
```
justinhynes, line 19: Nit: Equivalent is mispelled in the description.

src/components/bulk-email-tool/BulkEmailBody.jsx
```diff
@@ -1,5 +1,44 @@
...
+            />
+          </Form.Label>
+          <Form.Control className="w-lg-50" />
+        </Form.Group>
+        <Form.Group controlId="emailBody">
+          <Form.Label>
+            <FormattedMessage
+              id="bulk.email.body.label"
+              defaultMessage="Body:"
+              description="Email Body label. Meant to have colon or equivilant punctuation."
```
justinhynes, line 29: Same nit here.

justinhynes (approved):
LGTM! Thanks for resourcing the strings and creating the follow up ticket for feature parity in the editor.

PR #9, opened 2022-01-10
chore: Add exception for non es5 transpiled dep

Tj-Tracy:
During our CI checks on PRs, we check to see if all code can be properly
transpiled into es5 to support IE11. We've since dropped support
officially for IE11, but the check is still present. A dependency of
TinyMCE is triggering an error in the is-es5 check.

Adding this webpack config will except the package from the check, to
prevent the CI from failing. Since we no longer support IE11, This check
is not needed, and ok to bypass.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1632,12 +1632,6 @@
             "picocolors": "^1.0.0"
           }
         },
-        "caniuse-lite": {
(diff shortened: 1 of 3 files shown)


justinhynes (approved):
👍 of trust!

PR #10, opened 2022-01-12
feat: Add historical email content data to bulk course email task manager

justinhynes:
[MICROBA-1621]

* Add table that displays info about previously sent bulk course email messages through our tool.
* Add modal to view the contents of previously sent messages.
* Extract strings to a dedicated `messages.js` file to help make the `BulkEmailContentHistory.jsx` component more readable.

Screenshots:
Before button is clicked and the table is displayed:
Screen Shot 2022-01-12 at 3 41 42 PM
Screen Shot 2022-01-12 at 3 41 55 PM
Modal when the "View Message" link is clicked
Screen Shot 2022-01-12 at 3 42 16 PM
Warning Alert with message when there is no data to display:
Screen Shot 2022-01-12 at 3 42 29 PM
Error Alert with message when we could not fetch our data:
Screen Shot 2022-01-12 at 3 43 07 PM

### DIFF
--- a/src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
+++ b/src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
@@ -1,41 +1,241 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
 import { useParams } from 'react-router-dom';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 
-import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import {
+  Alert, Button, DataTable, Modal,
+} from '@edx/paragon';
+import messages from './messages';
 import { getSentEmailHistory } from './api';
 
-export default function BulkEmailContentHistory() {
+export function BulkEmailContentHistory({ intl }) {
   const { courseId } = useParams();
 
-  const [emailHistoryData, setEmailHistoryData] = useState(); // eslint-disable-line no-unused-vars
+  const [emailHistoryData, setEmailHistoryData] = useState();
+  const [errorRetrievingData, setErrorRetrievingData] = useState(false);
+  const [showHistoricalEmailContentTable, setShowHistoricalEmailContentTable] = useState(false);
+  const [isMessageModalOpen, setIsMessageModalOpen] = useState(false);
+  const [messageContent, setMessageContent] = useState();
 
-  useEffect(() => {
-    async function fetchSentEmailHistoryData() {
-      const data = await getSentEmailHistory(courseId);
+  /**
+   * Async function that makes a REST API call to retrieve historical email message data sent by the bulk course email
+   * tool from edx-platform.
+   */
+  async function fetchSentEmailHistoryData() {
+    let data = null;
+    try {
+      data = await getSentEmailHistory(courseId);
+    } catch (error) {
+      setErrorRetrievingData(true);
+    }
+
+    if (data) {
       const { emails } = data;
       setEmailHistoryData(emails);
+      setShowHistoricalEmailContentTable(true);
     }
-    fetchSentEmailHistoryData();
-  }, []);
+  }
+
+  /**
+   * This function is responsible for setting the current `messageContent` state data. This will be the contents of a
+   * previously sent email message from the bulk course email tool. This also toggles a modal to be visible to display
+   * the message contents to the end user.
+   */
+  const onViewMessageClick = (tableData) => {
+    setMessageContent(tableData);
+    setIsMessageModalOpen(true);
+  };
+
+  /**
+   * Render function for the email content history table. If an error occurs while attempting to fetch data from
+   * edx-platform we will render this error instead of the table.
+   */
+  const renderError = () => (
+    <div>
+      <Alert variant="danger">
+        <p className="font-weight-bold">
+          {intl.formatMessage(messages.errorFetchingData)}
+        </p>
+      </Alert>
+    </div>
+  );
+
+  /**
+   * Render function for the email content history table. If there is no data to display in our table we will render
+   * this informative message instead.
+   */
+  const renderEmpty = () => (
+    <div className="pt-1">
+      <Alert variant="warning">
+        <p className="font-weight-bold">
+          {intl.formatMessage(messages.noEmailData)}
+        </p>
+      </Alert>
+    </div>
+  );
+
+  /**
+   * Renders a modal that will display the contents of a single historical email message sent via the bulk course email
+   * tool to a user.
+   */
+  const renderMessageModal = () => (
+    <div>
+      <Modal
+        open={isMessageModalOpen}
+        title=""
+        body={(
+          <div>
+            <div className="d-flex flex-row">
+              <p>
+                {intl.formatMessage(messages.modalMessageSubject)}
+              </p>
+              <p className="pl-2">
+                {messageContent.subject}
+              </p>
+            </div>
+            <div className="d-flex flex-row">
+              <p>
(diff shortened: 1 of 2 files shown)


justinhynes:
Couple thoughts...

Is the `View Message` column header and the `View Message` button too redundant? This is a change from how this works today. In the current implementation clicking on the subject of the message would open the modal. I thought using the `additionalColumns` feature of the Paragon DataTable was a much quicker solution. 

Screenshot for comparison:
image

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
```diff
@@ -1,41 +1,241 @@
...
+} from '@edx/paragon';
+import messages from './messages';
 import { getSentEmailHistory } from './api';
 
-export default function BulkEmailContentHistory() {
+export function BulkEmailContentHistory({ intl }) {
   const { courseId } = useParams();
 
-  const [emailHistoryData, setEmailHistoryData] = useState(); // eslint-disable-line no-unused-vars
+  const [emailHistoryData, setEmailHistoryData] = useState();
```
hurtstotouchfire, line 15: @Tj-Tracy Justin said you could explain this state handling pattern to me 🙏

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
```diff
@@ -1,41 +1,241 @@
...
+              <p className="pl-2">
+                {messageContent.sent_to}
+              </p>
+            </div>
+            <hr className="py-2" />
+            <div>
+              <p>
+                {intl.formatMessage(messages.modalMessageBody)}
+              </p>
+              <div dangerouslySetInnerHTML={{ __html: messageContent.email.html_message }} />
```
hurtstotouchfire, line 126: WOW. What is this?

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
```diff
@@ -1,41 +1,241 @@
...
+
+  /**
+   * Render function for the email content history table. This function is responsible for displaying data inside of
+   * the table when the `Show Sent Email History` button is pressed on the page.
+   */
+  const renderTable = () => {
+    // Do a little data manipulation to make it easier to display what we want in the table. Pull the email subject out
+    // of the email data. Transforms the `sent_to` array to a string for easier display in our table.
+    const tableData = emailHistoryData.map((item) => ({
+      ...item,
```
hurtstotouchfire, line 143: What other key value pairs come out of this spread? What is the item arg an instance of? Is it documented somewhere? (fyi using jsdoc would potentially answer some of these questions)

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
justinhynes: This is to display the message HTML that we received from our API call. Without doing this, it shows the text with the markup ("<p>message</p>" rather than "message" (which we don't do today).

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
justinhynes: A link to some documentation that may be helpful as well: https://reactjs.org/docs/hooks-state.html

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
justinhynes: The `item` is referring to the individual object instance I am interacting with as I am looping through the `emailHistoryData` object. The `emailHistoryData` is an array of  previously sent email messages we got back from our REST API call to edx-platform. I needed to do some data transformation to make things fit the way the Paragon `DataTable` component wants.

We want to display the subject of the message in our data table but this data was buried a level deep in an inner object. When we define the columns of a DataTable, it wants an `accessor` key to get to the data in our objects. This piece of code is pulling that `subject` field out of the inner object and copying it to the top level of the object so it can be displayed in the table.

Before transformation it looks something like...
```
{ 
	created: "Dec 07, 2021 at 18:06 UTC"
	email: {subject: 'asdasdas', html_message: '<p>sfafasfdasd</p>\n<style></style>', id: '50'}
	number_sent: "1 sent"
	requester: "edx"
	sent_to: (2) ['Myself', 'All students']
}
```
and after the transformation it looks like...
```
{ 
	created: "Dec 07, 2021 at 18:06 UTC"
	email: {subject: 'asdasdas', html_message: '<p>sfafasfdasd</p>\n<style></style>', id: '50'}
	number_sent: "1 sent"
	requester: "edx"
	sent_to: "Myself, All students"
	subject: "asdasdas"
}
```

(I also fiddled with the `sent_to` data and transformed it from an array to a String. This was to fix another UI bug in the table. Keeping it an array will display `MyselfAll Students` (no spaces between entries) in the column. 

Hope this makes sense! Let me know if you have further questions.

src/components/bulk-email-tool/bulk-email-task-manager/messages.js
```diff
@@ -0,0 +1,78 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
```
justinhynes, line 1: I extracted the messages to a separate file with the addition of the tables and modal. I felt that it made the Component files a bit easier to read. I saw that this is a common practice in the learning MFE (and remember seeing it in other MFE's as well).

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
Tj-Tracy: functional components with hooks are how we tend to set up components now, as opposed to class components. `useState` returns a piece of state and the function needed to set the state for it.

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
Tj-Tracy: This is unfortunately the only way to do this. That's why the prop is named that way. If we want to keep the same features we have on the instructor tool, this is the way to do it.

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
```diff
@@ -1,41 +1,241 @@
...
+  const onViewMessageClick = (tableData) => {
+    setMessageContent(tableData);
+    setIsMessageModalOpen(true);
+  };
+
+  /**
+   * Render function for the email content history table. If an error occurs while attempting to fetch data from
+   * edx-platform we will render this error instead of the table.
+   */
+  const renderError = () => (
```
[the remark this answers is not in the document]
justinhynes: What is the best way to co-locate a reusable component like this? Do we have a pattern I should follow?

justinhynes:
@Tj-Tracy Sure. I was originally planning on adding tests last (after I implemented the tables in the other components) but I can swap this around. I'll update this PR with some tests and request a re-review.

Tj-Tracy (approved):
Unblocking because the testing suite needs some setup

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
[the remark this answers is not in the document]
Tj-Tracy: since I think this is something that would only be used by the `bulk-email-task-manager`, just as a single component in that folder that can be pulled in by the `contentHistory` and `pendingTasks` to display errors/empty state.

PR #12, opened 2022-01-19
feat: Implement task history and pending task tables

justinhynes:
[MICROBA-1621]
* Add a table to the BulkEmailTaskHistory component of the task manager.
* Add a table to the BulkEmailPendingTasks component of the task manager.
* Extract messages to messages.js.
* Add custom hook used to make REST API calls to edx-platform on a regular cadence (every 30 seconds).
* Use the new hook in BulkEmailPendingTasks to retrieve pending instructor tasks for display in a table.

Screenshots:
**Pending Tasks table**
image

**Pending Tasks table (empty)**
image

**Email Task History**
image

**Email Task History (empty)**
image

### DIFF
--- a/src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailPendingTasks.jsx
+++ b/src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailPendingTasks.jsx
@@ -1,32 +1,91 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 import { useParams } from 'react-router-dom';
 
-import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { getInstructorTasks } from './api';
+import messages from './messages';
+import useInterval from '../../../utils/useInterval';
+import BulkEmailTaskManagerTable from './BulkEmailHistoryTable';
 
-export default function BulkEmailPendingTasks() {
+export function BulkEmailPendingTasks({ intl }) {
   const { courseId } = useParams();
 
-  const [instructorTaskData, setInstructorTaskData] = useState(); // eslint-disable-line no-unused-vars
+  const [instructorTaskData, setInstructorTaskData] = useState();
+  const [errorRetrievingData, setErrorRetrievingData] = useState(false);
 
-  useEffect(() => {
+  /**
+   * We use a custom hook (`useInterval`) here to setup a timer that will refresh the pending instructor task data
+   * displayed in the table of this component.
+   */
+  useInterval(() => {
     async function fetchPendingInstructorTasksData() {
       const data = await getInstructorTasks(courseId);
       const { tasks } = data;
       setInstructorTaskData(tasks);
     }
-    fetchPendingInstructorTasksData();
-  }, []);
+
+    try {
+      fetchPendingInstructorTasksData();
+    } catch (error) {
+      setErrorRetrievingData(true);
+    }
+  }, 30000);
+
+  const tableColumns = [
+    {
+      Header: `${intl.formatMessage(messages.taskHistoryTableColumnHeaderTaskType)}`,
+      accessor: 'task_type',
+    },
+    {
+      Header: `${intl.formatMessage(messages.taskHistoryTableColumnHeaderTaskInputs)}`,
+      accessor: 'task_input',
+    },
+    {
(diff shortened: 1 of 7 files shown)


src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailPendingTasks.jsx
```diff
@@ -1,32 +1,89 @@
...
       const data = await getInstructorTasks(courseId);
       const { tasks } = data;
       setInstructorTaskData(tasks);
     }
-    fetchPendingInstructorTasksData();
-  }, []);
+    try {
+      fetchPendingInstructorTasksData();
+    } catch (error) {
+      // do nothing
```
justinhynes, line 24: I didn't want to just console log the error here but couldn't think of anything else we would want to do. If we can't fetch/refresh the data then we don't want to do reset it or anything, I think we just keep whatever our current results are and keep the table state as is.

I'd be swallowing the error here but I don't want the page to crash if we can't get results back from the API for whatever reason.

src/components/bulk-email-tool/bulk-email-task-manager/messages.js
```diff
@@ -2,7 +2,7 @@ import { defineMessages } from '@edx/frontend-platform/i18n';
 
 const messages = defineMessages({
   /* BulkEmailContentHistory.jsx Messages */
-  errorFetchingData: {
+  errorFetchingEmailHistoryData: {
```
justinhynes, line 5: Updated the name to be more specific since there are multiple date sets we are fetching for 1 of 3 tables.

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailPendingTasks.jsx
Tj-Tracy: You could use the paragon `Form.Control.Feedback` component for this and create a generic error message to give feedback to the user that something failed. If you use it together with `Form.Group` you shouldn't need any additional styling: https://paragon-edx.netlify.app/components/form/form-group/

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailTaskHistory.jsx
```diff
@@ -1,41 +1,149 @@
...
+          {intl.formatMessage(messages.emailTaskHistoryTableSectionButtonHeader)}
         </p>
-        <button type="button" className="btn btn-outline-primary mb-2">
-          <FormattedMessage
-            id="bulk.email.view.task.history.button"
-            defaultMessage="Show Task Email History"
-            description="Button that displays a table with historical bulk email task data for a course-run"
-          />
-        </button>
+        <Button variant="outline-primary" className="btn btn-outline-primary mb-2" onClick={async () => { await fetchEmailTaskHistoryData(); }}>
```
Tj-Tracy, line 136: I know I didn't mention this on the last PR, but it would be great if we could give feedback to the user in the form of a loading state. Either by using `StatefulButton` from paragon or just a general "disable button and show spinner" while waiting for the api to respond. Since the current version doesn't do this, I'll leave it up to you if you find it necessary (or as a fast-follow improvement)

https://paragon-edx.netlify.app/components/statefulbutton/

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailTaskHistory.jsx
```diff
@@ -1,41 +1,149 @@
...
     }
-    fetchEmailTaskHistoryData();
-  }, []);
+  }
+
+  /**
+   * Render function for the email task history table. If an error occurs while attempting to fetch data from
+   * edx-platform we will render this error instead of the table.
+   */
+  const renderError = () => (
```
Tj-Tracy, line 41: I think `renderError` and `renderEmpty` could be the same function. Just pass the error level and the message when rendering.

Tj-Tracy (approved):
A few notes but overall LGTM 👍

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailTaskHistory.jsx
justinhynes: Oh, this is a good idea and I'll take a look at the documentation! I think this would be a good addition overall.

Tj-Tracy (approved):
👍 Awesome job with this!

PR #17, opened 2022-02-07
feat: add features to text editor

Tj-Tracy:
image

### DIFF
--- a/src/components/bulk-email-tool/text-editor/TextEditor.jsx
+++ b/src/components/bulk-email-tool/text-editor/TextEditor.jsx
@@ -13,6 +13,7 @@ import 'tinymce/plugins/emoticons/js/emojis';
 import 'tinymce/plugins/link';
 import 'tinymce/plugins/lists';
 import 'tinymce/plugins/table';
+import 'tinymce/plugins/image';
 import '@edx/tinymce-language-selector';
 
 import contentUiCss from 'tinymce/skins/ui/oxide/content.css';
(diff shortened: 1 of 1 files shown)


src/components/bulk-email-tool/text-editor/TextEditor.jsx
```diff
@@ -30,12 +31,13 @@ export default function TextEditor(props) {
         selector: 'textarea#editor',
         height: 600,
         branding: false,
-        plugins: 'advlist code emoticons link lists table',
-        toolbar: 'bold italic | bullist numlist | link emoticons',
+        plugins: 'advlist code emoticons link lists table image language',
+        toolbar: 'formatselect fontselect bold italic underline forecolor | code bullist numlist alignlef aligncenter alignright alignjustify indent | link emoticons image | language',
```
justinhynes, line 35: A couple of thoughts...

1. In the original editor there is a "dedent"/`decrease indentation` option next to the the indent/"increase indentation" button. Can we add it?

2. In the original editor there is a button labeled `HTML` that allows direct manipulation of the HTML of the message. From a quick glance at the picture, I don't think I see this functionality. Is this still supported?

3. In the original editor there is a "remove link" option. Is that still supported? Should we add it?

4. What is "browser default language"? I don't know if I noticed that before. Should it be added? 

5. In the original editor there is a "blockquote" option. Is this still supported? Should we add it?

justinhynes (changes requested):
Couple of questions before approval

src/components/bulk-email-tool/text-editor/TextEditor.jsx
Tj-Tracy: tl;dr some buttons no longer exist on the toolbar and are now in context or sub menus due to version differences between here and platform. But, all the functionality is there. I realize the UI changes are not ideal, but I didn't want to spend extra time making custom buttons unless we decide its necessary.

1. This is the one tool I could not find, I don't think it still exists in this version. But, you can just press backspace, so to go back a step (or undo) so I didn't think it was that important.

2. there is an html button in the in the "tools" submenu 
image

3. this is now in a context menu.
image


4. Thats an edx made plugin that an instructor can use to denote a language in an email is not the same as a browser default. Its probably not going to get used, but is a nice-to-have from an i18n standpoint.

5. Block quotes are supposed to be there I just missed it by accident 😅 I've added it
image

justinhynes (approved):
Thanks for getting this squared away!

PR #18, opened 2022-02-08
fix: [MICROBA-1703] Add missing form instructions

Tj-Tracy:
Add/move instructions for submitting the email form.
Also fixes a sneaky bug when selecting cohorts.
image

### DIFF
--- a/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
+++ b/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
@@ -101,18 +101,11 @@ export default function BulkEmailForm(props) {
                 }}
               />
             </p>
-            <ul>
+            <ul className="list-unstyled">
               {selectedRecipients.map((group) => (
                 <li key={group}>{group}</li>
               ))}
             </ul>
-            <p>
-              <FormattedMessage
-                id="bulk.email.task.alert.warning"
-                defaultMessage="CAUTION! When you select Send Email, your email message is added to the queue for sending, and cannot be cancelled."
-                description="Warns the user in an alert that the email may not be immediately sent out to users."
-              />
-            </p>
           </>
         )}
         close={(event) => {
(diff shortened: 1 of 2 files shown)


src/components/bulk-email-tool/bulk-email-form/BulkEmailRecipient.jsx
```diff
@@ -82,7 +82,6 @@ export default function BulkEmailRecipient(props) {
               <Form.Checkbox
                 key={cohort}
                 value={`cohort:${cohort}`}
-                disabled={selectedGroups.find((group) => group === (DEFAULT_GROUPS.AUDIT || DEFAULT_GROUPS.VERIFIED))}
```
justinhynes, line 85: Is this the sneaky bugfix mentioned?

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -167,6 +160,17 @@ export default function BulkEmailForm(props) {
...
+          <p>
+            We recommend sending learners no more than one email message per week. Before you send your email, review
+            the text carefully and send it to yourself first, so that you can preview the formatting and make sure
+            embedded images and links work correctly.
+          </p>
+          <p>
+            <strong>CAUTION!</strong> When you select Send Email, your email message is added to the queue for sending,
+            and cannot be cancelled.
+          </p>
+        </div>
```
justinhynes, line 173: We should put this in a <FormattedMessage> to support i18n.

src/components/bulk-email-tool/bulk-email-form/BulkEmailRecipient.jsx
Tj-Tracy: yes - cohorts were being disabled when they weren't supposed to be because I forgot to delete this line in the copy paste. I just happen to notice it and it was a one-line fix.

justinhynes (approved):
A request to fix some hard-coded English but otherwise looks good! I don't need a re-review.

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
Tj-Tracy: Oh you think I would learn this lesson by now but here we are.

PR #22, opened 2022-02-28
fix: [MICROBA-1709] Fix cohort display bug

Tj-Tracy:
There was a small issue with the cohorts looking a little weird in the
checkbox list. This even out the list and forces each item to be the
same size.

## 2 cohorts
Desktop:
image

Mobile:
image

## A bunch of cohorts
Desktop:
image

Mobile:
image

### DIFF
--- a/src/components/bulk-email-tool/bulk-email-form/bulk-email-recipient/bulkEmailRecepient.scss
+++ b/src/components/bulk-email-tool/bulk-email-form/bulk-email-recipient/bulkEmailRecepient.scss
@@ -1,0 +1,9 @@
+.recipient-groups {
+  > div {
+    min-width: 50%;
+    padding-right: 0.5rem;
+    input {
+      padding: 0.5rem !important;
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


src/components/bulk-email-tool/bulk-email-form/bulk-email-recipient/bulkEmailRecepient.scss
```diff
@@ -0,0 +1,9 @@
+.recipient-groups {
+  > div {
+    min-width: 50%;
+    padding-right: 0.5rem;
+    input {
+      padding: 0.5rem !important;
+    }
+  }
+}
```
justinhynes, line 9: I am curious if I am reading this correctly. Still shaky on this and haven't seen this syntax too many times. This is overwriting the styling of the div(s?) in the `Form.CheckboxSet` component (sub-component?)? 

We do it this way with a selector (is that what the `>` is called?)? Is this targeting all divs in the component that uses this piece of CSS? Or just the first?

justinhynes (approved):
LGTM

src/components/bulk-email-tool/bulk-email-form/bulk-email-recipient/bulkEmailRecepient.scss
Tj-Tracy: to be honest, I have no clue the proper name 😅. the chevron denotes "all direct children of parent"; so in this case every `div` that is a child of this class gets the styles. so `recipient-groups > div` gets them but not `recipient-groups > div > div`.

PR #24, opened 2022-03-08
fix: Update webpack config to be less strict

justinhynes:
[MICROBA-1743]

Fixes an issue where we couldn't override the header and footer with the branded versiom within the Comms MFE because some of the dependencies of the branded components are not currently Webpack5 compliant. Adds a rule to make webpack5 less strict with imports that do not include a file extension.

This also includes a change so that the Comms MFE uses the Open edX branded colors/styles by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,9 +35,9 @@
     "url": "https://github.com/edx/frontend-app-communications/issues"
   },
   "dependencies": {
-    "@edx/brand": "npm:@edx/brand-edx.org@^2.0.3",
+    "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
     "@edx/frontend-component-footer": "10.1.6",
-    "@edx/frontend-component-header": "2.3.0",
+    "@edx/frontend-component-header": "^2.4.2",
     "@edx/frontend-platform": "1.12.7",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -35,9 +35,9 @@
     "url": "https://github.com/edx/frontend-app-communications/issues"
   },
   "dependencies": {
-    "@edx/brand": "npm:@edx/brand-edx.org@^2.0.3",
+    "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
```
justinhynes, line 38: I thought we would want the Open edX branded version of the package by default. We will override this dependency in an upcoming `edx-internal` PR following these changes.

package.json
```diff
@@ -35,9 +35,9 @@
...
     "@edx/frontend-component-footer": "10.1.6",
-    "@edx/frontend-component-header": "2.3.0",
+    "@edx/frontend-component-header": "^2.4.2",
```
justinhynes, line 40: Since I was in here I figured we should use the latest version of the (unbranded) frontend-component-header.

Tj-Tracy (approved):

PR #25, opened 2022-03-09
fix: [Microba-1741] Fix email image upload

Tj-Tracy:
Image uploads were broken for 2 reasons on stage:
  1. Our bleaching on the server was too strong. it was pulling out
     attributes inside of the image tags that allowed base64 encoding to
     work properly.
  2. The paste_data_images needs to be set to true to allow for dragging
     and dropping of images into the editor

Since bleach has been rolled back for now, adding this setting should
allow for images to be properly uploaded.


justinhynes (approved):
LGTM!

PR #27, opened 2022-03-11
feat: suggested UI improvements from bugbash

justinhynes:
[MICROBA-1743]

* Use `LearningHeader` component which can display the course information (title, org, number) in the header to the learner (matching existing design).
* Refactor components to introduce a `PageContainer` component in order to support displaying course information in header.
* Refactor the BulkEmailTool component to take advantage of the new course metadata context provider.
* Add tests for new `PageContainer` component.
* Fix existing tests due to refactor.
* Add "Email" label above recipient selection check boxes to match existing design

Before:
image

After:
image

P.S. Couple of additional notes: 
* I believe the logo size to be a local environment thing, shouldn't be this large on Stage or Prod. 
* Other visual header changes come from using the `LearningHeader` over the plain ole regular `Header` component. The `LearningHeader` already had support for displaying course information as part of the component.


src/components/page-container/data/api.js
```diff
@@ -1,14 +1,8 @@
 import { getConfig } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 
-const InstructorApiBaseUrl = `${getConfig().LMS_BASE_URL}/api/instructor/v1`;
 const courseHomeBaseUrl = `${getConfig().LMS_BASE_URL}/api/course_home/v1/course_metadata`;
 
-export default async function getInstructorData(courseId) {
```
justinhynes, line 7: I searched the codebase and this must have been a leftover from an early part of our implementation. It wasn't being called anywhere.

src/components/page-container/PageContainer.jsx
```diff
@@ -0,0 +1,87 @@
...
+export const CourseMetadataContext = React.createContext();
+
+export default function PageContainer(props) {
+  const { children } = props;
+  const { courseId } = useParams();
+
+  const [courseMetadata, setCourseMetadata] = useState();
+
+  useEffect(() => {
+    async function fetchCourseMetadata() {
```
Tj-Tracy, line 20: Eventually, we may want to extract this as a separate function and pass it to the `PageContainer` as a prop, so that the component is more generic. This way, if a new page down the line needs more/different data, we don't have to modify this component.

For now, I'm happy with the way its done here. I don't know if we need to refactor this until that case comes up.

Tj-Tracy (approved):
Great work 👍 One food-for-thought comment, But I decided we might not need to worry about it for the time being.

src/components/page-container/PageContainer.jsx
justinhynes: Talked with TJ offline about this. I'm going to leave it as-is for now so we can continue to focus on other release of the MFE. We may need to come back around to this in the future when/if we add new functionality to the Comms MFE. For now, this shouldn't hurt staying the way it is.

PR #29, opened 2022-03-15
fix: [MICROBA-1760] code block on toolbar

Tj-Tracy:
The new plugin for code blocks is in the sameplace as the old one, and the HTML editor is back in the old spot as well.

image


justinhynes (approved):
Good eyes and thanks for catching this!

PR #34, opened 2022-04-15
fix: [MICROBA-1799] fix bulk email tool on safari

Tj-Tracy:
Course teams were having issues sending bulk emails to themselves and
students. This was caused by two problems.

1. The language selector tool was failing because of no default language
   settings in safari is possible
2. The translated string for the "continue" button on the submit modal
   was messing with the markup and causing the event to POST the email
to not properly fire.

To fix the language issue for now, we are disabling the language
selector plugin. To fix the markup issue, we are forcing the string to
render in a fragment to remove the additional span, allowing the event
to fire no matter where the user clicks the button.

### DIFF
--- a/src/components/bulk-email-tool/bulk-email-form/TaskAlertModal.jsx
+++ b/src/components/bulk-email-tool/bulk-email-form/TaskAlertModal.jsx
@@ -35,7 +35,14 @@ function TaskAlertModal(props) {
               id="bulk.email.form.recipients.Contine"
               defaultMessage="Continue"
               description="Continue button for the task alert"
-            />
+            >
+              { // FormattedMessage wraps the translated string in a <span/> by default. This was
+                // causing strange click event target issues in safari. To solve this, we want to
+                // wrap the string in a fragment instead of a span, so that the whole button considered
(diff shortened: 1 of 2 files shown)


src/components/bulk-email-tool/bulk-email-form/TaskAlertModal.jsx
```diff
@@ -35,7 +35,9 @@ function TaskAlertModal(props) {
               id="bulk.email.form.recipients.Contine"
               defaultMessage="Continue"
               description="Continue button for the task alert"
-            />
+            >
+              {msg => <>{msg}</>}
```
justinhynes, line 39: nit: This looks a bit cryptic. Is there any way we'd be able to add a comment here/nearby that explains why we have to do this? I'm not sure I would remember why we did this in a few months time.

justinhynes (approved):
Thanks so much for your help on this one!

src/components/bulk-email-tool/bulk-email-form/TaskAlertModal.jsx
Tj-Tracy: absolutely! comment added!

PR #35, opened 2022-04-21
chore: update edx frontend packages

Tj-Tracy:


src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
```diff
@@ -11,7 +11,7 @@ import messages from './messages';
 import { getSentEmailHistory } from './data/api';
 import BulkEmailTaskManagerTable from './BulkEmailHistoryTable';
 
-export function BulkEmailContentHistory({ intl, copyTextToEditor }) {
```
Tj-Tracy, line 14: All these import changes are just needed for linting. Since these are default exported anyway the export here is unneeded.

justinhynes (approved):

PR #36, opened 2022-04-25
feat: [MICROBA-1528] add schedule email UI

Tj-Tracy:
This handles a few things around the scheduled email UI. This includes:
1. Adding the schedule email UI date/time picker.
2. Adds states to the submit button for scheduling emails.
3. Drys out intl code and some submitting states.
4. Matches the email form UI to mocks.

This however does not include:
1. A table to show scheduled emails. Scheduled emails at the moment are
   displayed in the "pending tasks" section.
2. Matching the tasks section to the mocks.

## Desktop
image
image
image
image

## Mobile
image
image
image


src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -115,127 +146,98 @@ export default function BulkEmailForm(props) {
...
-                id="bulk.email.form.error"
-                defaultMessage="An error occured while attempting to send the email."
-                description="An Error message located under the submit button for the email form. Visible only on a failure."
-              />
-            </Form.Control.Feedback>
-          )}
-          {(emailFormStatus === FORM_SUBMIT_STATES.COMPLETED_DEFAULT
-            || emailFormStatus === FORM_SUBMIT_STATES.COMPLETE) && (
-            <Form.Control.Feedback className="px-3" hasIcon={false} type="valid">
-              <FormattedMessage
```
Tj-Tracy, line 232: This messaging is not in the new mocks, and so I've removed it. Theres no longer a success message when an email is successfully sent, besides the messaging on the submit button.

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -70,6 +94,10 @@ export default function BulkEmailForm(props) {
       emailData.append('send_to', JSON.stringify(selectedRecipients));
       emailData.append('subject', subject);
       emailData.append('message', editorRef.current.getContent());
+      if (isScheduled) {
+        const { scheduleDate, scheduleTime } = dateTime;
+        emailData.append('schedule', new Date(`${scheduleDate} ${scheduleTime}`).toISOString());
+      }
```
justinhynes, line 100: Can we also make sure to capture and pass along the browser timezone? This will be used as a fallback option if the user doesn't have a "preferred timezone" set in their account settings.

The backend is expecting checking for a key named `browser_timezone` in the data: https://github.com/openedx/edx-platform/blob/f2582c4e9e6664fd77ee8cfe648b40558962064a/lms/djangoapps/instructor/views/api.py#L2709.

justinhynes: The time zone only really matters to scheduled requests (at this time) so I think it's fine to skip sending that data on non-scheduled requests.

src/components/bulk-email-tool/bulk-email-form/messages.js
```diff
@@ -0,0 +1,89 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
```
justinhynes, line 1: Thanks for doing this! I think it makes the component source code much easier to read with these resourced in a dedicated file.

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -47,19 +64,26 @@ export default function BulkEmailForm(props) {
   };
 
   const onSubjectChange = (event) => setSubject(event.target.value);
+  const onDateTimeChange = (event) => {
+    const next = { [event.target.name]: event.target.value };
+    setDateTime((prev) => ({ ...prev, ...next }));
+  };
```
justinhynes, line 70: I don't think I'm following what this is for? Sorry to be daft, do you mind explaining what the purpose of `onDateTimeChange` is for?

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -47,19 +64,26 @@ export default function BulkEmailForm(props) {
...
 
   const validateEmailForm = () => {
     const subjectValid = subject.length !== 0;
     const bodyValid = editorRef.current.getContent().length !== 0;
     const recipientsValid = selectedRecipients.length !== 0;
+    const { scheduleDate, scheduleTime } = dateTime;
+    const scheduleValid = isScheduled ? (scheduleDate && scheduleTime) !== undefined : true;
```
justinhynes, line 77: I found this a little hard to read/grok at first. Just to make sure I _am_ reading it right... `isScheduled` is the condition and then we have a second conditional after the `?` --> (`(scheduleDate && scheduleTime) !== undefined`) that is determining the value if `isScheduled` is true? otherwise the conditional is false (if `isScheduled` is false?)?

I'd consider breaking it up into multiple lines if possible to make it easier to read, or trying to keep the conditional on the left side of the `?`.

Feel free to disagree and keep as-is. Just wanted to mention that I didn't find it entirely easy to read.

src/index.scss
```diff
@@ -1,8 +1,7 @@
-@import "@edx/brand/paragon/fonts.scss";
-@import "@edx/brand/paragon/variables.scss";
-@import "@edx/paragon/scss/core/core.scss";
-@import "@edx/brand/paragon/overrides.scss";
-
+@import "~@edx/brand/paragon/fonts";
```
Tj-Tracy, line 1: more house cleaning - these were improperly imported.

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
Tj-Tracy: yeah this is one of those things I built in pieces and didnt take the time to reread my draft. I can make this more stepped.

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
Tj-Tracy: I'm using a single piece of state and a single `onChange` function to capture the date and time (they come from separate inputs). This could have been inline on the component, but I extracted it out because I'm using it in two places. `event.target` will either be the date _or_ the time input when this executes.

the reason why I'm defining `next` here is because setting state is async, and the event target may no longer be the present when `setDateTime()` is executing. 

in practice:
- The user either inputs the date or the time in one of the input fields
- `onDateTimeChange` executes, capture the value in the `next` object
- we spread the new value in `next` over the previous object in the `dateTime` state. This preserves any value not being overwritten (like if the date is already present and the user edits the time).

src/components/bulk-email-tool/bulk-email-task-manager/test/BulkEmailContentHistory.test.jsx
```diff
@@ -16,6 +16,9 @@ jest.mock('../data/api', () => ({
 
 describe('BulkEmailContentHistory component', () => {
   beforeEach(() => jest.resetModules());
+  beforeAll(async () => {
```
Tj-Tracy, line 19: I needed to change a little bit of logic around the `setupTest` tools we use. To prevent tests from breaking because of `InltlProvder`, we run a setup function before all the tests to mock certain things out.

Tj-Tracy: This does mean I had to touch every test file, but for now we don't have that many.

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -115,127 +153,102 @@ export default function BulkEmailForm(props) {
...
-              <FormattedMessage id="bulk.email.instructions.caution" defaultMessage="Caution!" />
-            </strong>
-            <FormattedMessage
-              id="bulk.email.instructions.caution.message"
-              defaultMessage=" When you select Send Email, your email message is added to the queue for sending,
-              and cannot be cancelled."
-              description="A warning about how emails are sent out to users"
-            />
+            <strong>{intl.formatMessage(messages.bulkEmailInstructionsCaution)}</strong>
+            {intl.formatMessage(messages.bulkEmailInstructionsCautionMessage)}
```
justinhynes, line 185: What should we do with this messaging if the email author wants to schedule their message? The context doesn't make sense because eventually we _will_ allow a message to be cancelled. 

Should this only display when scheduling is disabled or the `Schedule this email...` checkbox isn't selected?

src/index.jsx
```diff
@@ -39,13 +39,16 @@ subscribe(APP_INIT_ERROR, (error) => {
...
-    mergeConfig({
-    }, 'CommuncationsAppConfig');
+  handlers: {
+    config: () => {
+      mergeConfig(
+        {
+          // MICROBA-1505: Remove this when we remove the flag from config
+          SCHEDULE_EMAIL_SECTION: process.env.SCHEDULE_EMAIL_SECTION || null,
+        },
+        'CommuncationsAppConfig',
```
justinhynes, line 49: Nit: Realized that `Communications` is mispelled haha

PR #42, opened 2022-05-17
refactor: [MICROBA-1506] refactor code to match ADR

Tj-Tracy:
This PR refactors some of the code around the context store to be more in line with the project organization ADR in this repo. Essentially, it splits the reducers and actions into slices used by the components that need them to prevent pollution of data in the store.

This PR also handles most of the refactor around the `BulkEmailTool` making use of the `BulkEmailToolProvider` in order to share data between components. This allows for better `copyToEditor` functionality, amongst other changes, as the email form now handles its state within the context store.

The Provider and its store is purposefully tied to the `BulkEmailTool` as to prevent any bleeding of state information between tools that may be added to the comms MFE in the future. 

---

This PR also adds the first iteration of the scheduled emails table. This table will allow for viewing, deleting, and editing emails in the future. For now, it only adds viewing. The viewing modal DOES support copy to editor functionality but it is NOT editing the original entry and WILL schedule a new email if submitted.

image

image


src/components/bulk-email-tool/bulk-email-task-manager/ViewEmailModal.jsx
```diff
@@ -1,12 +1,15 @@
...
 import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import messages from './messages';
+import { BulkEmailContext } from '../bulk-email-context';
+import { copyToEditor } from '../bulk-email-form/data/actions';
 
 function ViewEmailModal({
-  intl, messageContent, isOpen, setModalOpen, copyTextToEditor,
+  intl, messageContent, isOpen, setModalOpen,
 }) {
+  const [, dispatch] = useContext(BulkEmailContext);
```
justinhynes, line 12: Is there a missing first argument? Or is this an errant comma? Or is this syntax ([, dispatch]) mean something I might not be familiar with?

src/components/bulk-email-tool/BulkEmailTool.jsx
```diff
@@ -7,29 +7,26 @@ import BulkEmailTaskManager from './bulk-email-task-manager/BulkEmailTaskManager
...
   return (
     <CourseMetadataContext.Consumer>
       {(courseMetadata) => (courseMetadata.isStaff ? (
         <div>
-          <NavigationTabs courseId={courseId} tabData={courseMetadata.tabs} />
-          <div>
-            <div className="row">
-              <BulkEmailForm courseId={courseId} cohorts={courseMetadata.cohorts} editorRef={textEditorRef} />
+          <BulkEmailProvider>
+            <NavigationTabs courseId={courseId} tabData={courseMetadata.tabs} />
```
justinhynes, line 20: When we were going over this PR synchronously, I thought you mentioned that we may want to pull NavigationTabs out and be alongside/above the BulkEmailProvider. Just thought I may want to capture this in a comment in case it is something we need to change!

src/components/bulk-email-tool/bulk-email-task-manager/ViewEmailModal.jsx
Tj-Tracy: Its meant to be similar to a tuple in python. It does seem a bit odd in javascript, but this component has no need to access the store directly so I left the first piece of the tuple blank.

src/components/bulk-email-tool/BulkEmailTool.jsx
Tj-Tracy: Thank you for capturing that! I went ahead and made the change.

src/components/bulk-email-tool/bulk-email-task-manager/bulk-email-scheduled-emails-table/BulkEmailScheduledEmailsTable.jsx
```diff
@@ -0,0 +1,117 @@
...
+  return (
+    <>
+      {viewModal.isOpen && (
+        <ViewEmailModal
+          isOpen={viewModal.isOpen}
+          setModalOpen={(open) => setViewModal({ isOpen: open })}
+          messageContent={viewModal.messageContent}
+        />
+      )}
+      <DataTable
```
justinhynes, line 66: Does the data table have some way to re-send the API call to refresh the contents? If not, should we use the same mechanism the pending task table uses today to refresh the data in the table at a regular interval?

src/components/bulk-email-tool/bulk-email-task-manager/ViewEmailModal.jsx
justinhynes: Thanks for the context!

justinhynes:
Nit: From the screenshot, this looks like it is sitting right on top of the Pending Tasks table. Should we add some space between these two components?

src/components/bulk-email-tool/bulk-email-task-manager/bulk-email-scheduled-emails-table/BulkEmailScheduledEmailsTable.jsx
Tj-Tracy: I can create one: https://paragon-openedx.netlify.app/components/datatable/#bulk-actions

I would prefer to allow the user to manually refresh the table. I can also have the submit email button refresh the table on a successfully scheduled email.

I can add it to this PR or add it to the coming work around this table.

Tj-Tracy:
I added `pb-4` to the table to fix the margin issue. This is the same as the other tables below the component. Ill update the top comment with a new screenshot, and add a screenshot of the error state.

src/components/bulk-email-tool/bulk-email-task-manager/bulk-email-scheduled-emails-table/BulkEmailScheduledEmailsTable.jsx
Tj-Tracy: Ive added this to the AC of the next subtask.

justinhynes: Thanks, TJ!

justinhynes (approved):

PR #43, opened 2022-05-31
fix: use `originalUserIsStaff` to gate bulk email tool over `isStaff`

justinhynes:
[MICROBA-1822][CR-4822]

* use `originalUserIsStaff` to gate bulk email tool access over `isStaff`

Additional Context

Access to this tool was originally gated by looking at the `isStaff` value from the CourseMetadata data returned to us from the LMS API. The user masquerading feature seems to have some interesting interactions with the Instructor Dashboard and we may be denying legitimate staff/admin/instructors access to the tool. Instead of using the value of `isStaff` we will now use `originalUserIsStaff` to gate access (and this follows how the Learning MFE gates content as well).


src/components/page-container/PageContainer.jsx
```diff
@@ -31,23 +31,23 @@ export default function PageContainer(props) {
...
-          isStaff: false,
+          originalUserIsStaff: false,
           tabs: [],
           cohorts: [],
         });
         return;
       }
 
       const {
-        org, number, title, tabs, is_staff: isStaff,
```
Tj-Tracy, line 42: While were changing this stuff, can we do this camel case conversion in `api.js` instead of the component itself? theres a `camelCaseObject` function you can import in `@edx/frontend-platform`.

Tj-Tracy (changes requested):

src/components/page-container/PageContainer.jsx
justinhynes: I'll look into this. Thanks!

Tj-Tracy (approved):

PR #44, opened 2022-06-14
feat: [MICROBA-1840] delete/edit scheduled emails

Tj-Tracy:
This PR adds delete and edit functionality to emails that are scheduled to be sent at a future date. The workflow for editing is as such:
- A user clicks edit on the table
- The contents of the email are copied to the editor, and the user makes edits
- The user is warned that this is editing an email, and that it will not create a new task, instead updating the selected one
- If they accept, the email is updated.

For delete, the user just hits the delete button, and its gone. There is now warning popup as of yet.

This PR also adds the toast component for success and errors when sending emails.

image
image
image
image

### DIFF
--- a/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
+++ b/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
@@ -1,7 +1,7 @@
 import React, { useContext, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import {
-  Form, Icon, StatefulButton, useCheckboxSetValues, useToggle,
+  Form, Icon, StatefulButton, Toast, useToggle,
 } from '@edx/paragon';
 import {
   SpinnerSimple, Cancel, Send, Event, Check,
@@ -10,7 +10,6 @@ import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import classNames from 'classnames';
 import { getConfig } from '@edx/frontend-platform';
 import TextEditor from '../text-editor/TextEditor';
-import { postBulkEmail } from './data/api';
 import BulkEmailRecipient from './bulk-email-recipient';
 import TaskAlertModal from './TaskAlertModal';
 import useTimeout from '../../../utils/useTimeout';
@@ -18,7 +17,16 @@ import useMobileResponsive from '../../../utils/useMobileResponsive';
 import ScheduleEmailForm from './ScheduleEmailForm';
 import messages from './messages';
 import { BulkEmailContext } from '../bulk-email-context';
-import { handleEditorChange } from './data/actions';
+import {
+  addRecipient,
+  clearDateTime,
+  clearEditor,
+  handleEditorChange,
+  removeRecipient,
+  setEditMode,
+} from './data/actions';
+import { editScheduledEmailThunk, postBulkEmailThunk } from './data/thunks';
+import { getScheduledBulkEmailThunk } from '../bulk-email-task-manager/bulk-email-scheduled-emails-table/data/thunks';
 
 export const FORM_SUBMIT_STATES = {
   DEFAULT: 'default',
@@ -26,7 +34,13 @@ export const FORM_SUBMIT_STATES = {
   COMPLETE: 'complete',
   COMPLETE_SCHEDULE: 'completed_schedule',
   SCHEDULE: 'schedule',
+  RESCHEDULE: 'reschedule',
   ERROR: 'error',
+};
+
+const FORM_ACTIONS = {
+  POST: 'POST',
+  PATCH: 'PATCH',
 };
 
 function BulkEmailForm(props) {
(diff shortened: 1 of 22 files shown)


src/components/bulk-email-tool/bulk-email-form/data/actions.js
```diff
@@ -5,10 +5,64 @@ export const handleEditorChange = (fieldName, fieldValue) => ({
...
+
+export const patchScheduledEmailComplete = () => ({
+  type: 'PATCH_COMPLETE',
+});
+
+export const patchScheduledEmailError = () => ({
+  type: 'PATCH_FAILURE',
+});
+
+export const postBulkEmail = () => ({
```
justinhynes, line 55: Same question here?

src/components/bulk-email-tool/bulk-email-form/data/api.js
```diff
@@ -1,8 +1,20 @@
...
+import { logError } from '@edx/frontend-platform/logging';
 
-export async function postBulkEmail(email, courseId) {
+export async function postBulkEmailInstructorTask(email, courseId) {
   const url = `${getConfig().LMS_BASE_URL}/courses/${courseId}/instructor/api/send_email`;
   return getAuthenticatedHttpClient().post(url, email);
 }
+
+export async function patchScheduledBulkEmailInstructorTask(emailData, courseId, taskId) {
+  const endpointUrl = `${getConfig().LMS_BASE_URL}/api/instructor_task/v1/schedules/${courseId}/bulk_email/${taskId}`;
```
justinhynes, line 12: Nit: Do you think it would be better to rename `taskId` to `scheduleId`? Since the item we are updating is the actual instructor task schedule, and not the instructor task. Or, maybe we could write a comment about what the expected value of `taskId` represents?

That might reduce potential confusion down the road.

src/components/bulk-email-tool/bulk-email-task-manager/bulk-email-scheduled-emails-table/data/api.js
```diff
@@ -13,3 +15,14 @@ export async function getScheduledBulkEmailData(courseId, page = 1) {
     throw new Error(error);
   }
 }
+
+export async function deleteScheduledBulkEmailInstructorTask(courseId, taskId) {
+  const endpointUrl = `${getConfig().LMS_BASE_URL}/api/instructor_task/v1/schedules/${courseId}/bulk_email/${taskId}`;
```
justinhynes, line 20: Nit: Same suggestion as above. Schedule we update `taskId` to be `scheduleId` or `taskScheduleId` to better indicate that we are trying to remove the _schedule_ instance?

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -40,26 +53,60 @@ function BulkEmailForm(props) {
...
     schedule: true,
   });
-  const [selectedRecipients, { add, remove }] = useCheckboxSetValues([]);
   const [isTaskAlertOpen, openTaskAlert, closeTaskAlert] = useToggle(false);
   const [isScheduled, toggleScheduled] = useState(false);
   const isMobile = useMobileResponsive();
 
-  const resetEmailForm = useTimeout(() => {
+  /**
+   * Since we are working with both and old and new API endpoint, the body for the POST
```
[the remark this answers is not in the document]
Tj-Tracy: changed

src/components/bulk-email-tool/bulk-email-form/data/actions.js
Tj-Tracy: fixed

src/components/bulk-email-tool/bulk-email-form/data/api.js
Tj-Tracy: changed

src/components/bulk-email-tool/bulk-email-form/test/BulkEmailForm.test.jsx
```diff
@@ -110,6 +122,36 @@ describe('bulk-email-form', () => {
...
+          isLoading: false,
+          errorRetrievingData: false,
+        },
+      }),
+    );
+    const submitButton = screen.getByText('Reschedule Email');
+    fireEvent.click(submitButton);
+    expect(
+      await screen.findByText(
+        'This will not create a new scheduled email task and instead overwrite the one currently selected. Do you want to overwrite this scheduled email?',
```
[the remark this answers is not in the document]
Tj-Tracy: I think if we update the text we want to update the test. I could just used the translation string itself (ie: `messages.messageId`), but we usually do it this way so people are more conscientious of copy/translation changes.

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -85,62 +135,95 @@ function BulkEmailForm(props) {
...
+      let response;
+      if (editor.editMode) {
+        const editedEmail = formatDataForFormAction(FORM_ACTIONS.PATCH);
+        response = await dispatch(editScheduledEmailThunk(editedEmail, courseId, editor.schedulingId));
+      } else {
+        const emailData = formatDataForFormAction(FORM_ACTIONS.POST);
+        response = await dispatch(postBulkEmailThunk(emailData, courseId));
       }
-      if (data.status === 200) {
+      if (response?.status === 200) {
```
justinhynes, line 148: For my own edification, what does `response?` mean here? Or, more specifically, what does the `?` do here?

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -85,62 +135,95 @@ function BulkEmailForm(props) {
...
   };
 
   useEffect(() => {
+    if (!!editor.scheduleDate || !!editor.scheduleTime) {
+      toggleScheduled(true);
+    }
     if (isScheduled) {
-      setEmailFormStatus(FORM_SUBMIT_STATES.SCHEDULE);
+      if (editor.editMode) {
+        setEmailFormStatus(FORM_SUBMIT_STATES.RESCHEDULE);
```
justinhynes, line 168: Once the email form has entered `editMode`, does the user have a way to cancel or back out?

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
Tj-Tracy: A fairly new Js thing called [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining). its the same as `response && response.status === 200`.

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
Tj-Tracy: unchecking the scheduling box, but other than that, no. I could add a cancel button? maybe next to the send button?

src/components/bulk-email-tool/bulk-email-form/test/BulkEmailForm.test.jsx
```diff
@@ -62,9 +74,8 @@ describe('bulk-email-form', () => {
     expect(await screen.findByText('Email Created')).toBeInTheDocument();
   });
   test('Shows Error on failed POST', async () => {
-    postBulkEmail.mockImplementation(() => {
-      throw Error('api-response-error');
-    });
+    const axiosMock = new MockAdapter(getAuthenticatedHttpClient());
```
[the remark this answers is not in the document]
Tj-Tracy: It depends on what you're trying to test, but for the most part, yes. if you don't have any logic present in the `api` function, mocking the whole module is probably fine too.

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
justinhynes: I think a cancel button may be more intuitive to the users of the tool. Think this is something we need to run past Gabe?

I'd hate to delay this any further, and the PR already has a lot of changes. Maybe it can be a fast followup?

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
justinhynes: Oh cool! Thanks for the link. :)

justinhynes (approved):

PR #49, opened 2022-06-27
feat: Add cancel button for editing an email

Tj-Tracy:
Right now, canceling edit mode is clunky. This adds a button to explicitly cancel the mode for the editor.

When editing:
image
Hover: 
image

### DIFF
--- a/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
+++ b/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
@@ -1,6 +1,7 @@
 import React, { useContext, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import {
+  Button,
   Form, Icon, StatefulButton, Toast, useToggle,
 } from '@edx/paragon';
 import {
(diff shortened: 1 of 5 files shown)


src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -290,10 +279,11 @@ function BulkEmailForm(props) {
           <div
             className={classNames('d-flex', {
               'mt-n4.5': !isScheduled && !isMobile,
-              'flex-row-reverse justify-content-between align-items-end': !isMobile,
+              'flex-row-reverse align-items-end': !isMobile,
               'border-top pt-2': isScheduled,
             })}
           >
+            {editor.editMode && <Button className="ml-2" variant="outline-brand" onClick={() => dispatch(clearEditor())}>Cancel</Button>}
```
justinhynes, line 286: Cancel looks like it's hard-coded english here. Can we resource this string?

justinhynes (approved):
LGTM (other than the comment about the hardcoded text for the `Cancel` button). Thanks TJ!

justinhynes:
Side note: I saw that the tests were failing for Node12, but we should be good to drop support for that (since we're deploying on Node16 now). 

There is a PR for this here: https://github.com/edx/frontend-app-communications/pull/32 but it looks like there may be a conflict. I think once this lands it would likely solve the test issues.

Tj-Tracy:
This is actually failing a test due to a data change. Fixing it now.

PR #55, opened 2022-07-20
feat: [MICROBA-1902] Confirm email deletion

Tj-Tracy:
This PR adds an alert style pop up to confirm with the user that they
are deleting an email from the scheduled email table intentionally.

image


src/components/bulk-email-tool/task-alert-modal/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './TaskAlertModal';
```
Tj-Tracy, line 1: I moved the `TaskAlertModal` to its own folder with its own public API because its now used in more places then just local to the `BulkEmailForm`

justinhynes (approved):
LGTM!

PR #56, opened 2022-07-22
fix: [MICROBA-1905] fix cancel button

Tj-Tracy:
The various form states were getting mixed up during error handling,
which lead to some strange situations where the cancel button was
appearing when it shouldnt have been. This PR centralizes all of the
form state changes into a useEffect hook to handle any and all state
changes throughout the form. This approach not only centralizes the
fragmented code, but prevents the state mixups that were happening
previously.

It also has the added benefit if down the line more state changes need
to be added, the changes are all happening in one place.


justinhynes (approved):
LGTM!

PR #71, opened 2022-08-18
fix: [APER-1936] Changes for a11y review

Tj-Tracy:
changes as per the a11y review shake out as follows:

> The SkipNav link (Skip to main content) should jump to the H1 or to a link just above the H1, which would show on focus (you could use the same style as the SkipNav link to do this).

The skipNav link now properly anchors to the `H1`:

image

---
> The <html> element needs a lang attribute

image

---
> The page needs a unique <title> (text used for the browser tab) in the <head>.

image
image

---
> The label Body above the TinyMCE editor has a for= attribute that points to a missing ID. I think that ID should be on the `<textarea>` just below it (not the `<div role="application">`). I'm not 100% sure on that but please add it to confirm.

As far as I can tell, this is done correctly, as the role is where the `TinyMCE` editor starts:
`The application role indicates to assistive technologies that an element and all of its children should be treated similar to a desktop application, and no traditional HTML interpretation techniques should be used. `

this was done my TinyMCE itself, and not by us, so I caution messing with it.

---
> In the alert where it says "To view all pending tasks, including email, visit Course Info..." that inline link should have the inline link styling, which includes an underline in all states.

image
image

---
> The page doesn't seem to have a `<main>` (though there is a `<header>` and `<footer>`. Could you add one where the blank div is now (first sibling after `<header>`)?

image

---
> In the email Task History table, there are a bunch of links with text of View Message. Links are supposed to be unique on the page. Could you append `<span class="sr-only"> X</span>` (where X is the message number) at the end of the links?  Note the leading space before X, so it reads correctly with TTS.

image

---
> Are you sure that's the latest standard footer ?  The social media icons' color looks old for some reason. Not sure though. But on that note, the `<footer>` has aria-label="edX Home" which is totally random (so please just remove that), but I have seen that particular bug in older instances of the `<footer>`.

These were changed in an `edx-internal` PR: https://github.com/edx/edx-internal/pull/7003

### DIFF
--- a/src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
+++ b/src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
@@ -108,8 +108,8 @@ function BulkEmailContentHistory({ intl }) {
         Header: '',
         Cell: ({ row }) => (
           <Button variant="link" className="px-1" onClick={() => onViewMessageClick(tableData[row.index])}>
-            <span className="sr-only"> {row.index}</span>
             {intl.formatMessage(messages.buttonViewMessage)}
+            <span className="sr-only">&nbsp;{row.index}</span>
           </Button>
         ),
       },
--- a/src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
+++ b/src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
@@ -108,6 +108,7 @@ function BulkEmailContentHistory({ intl }) {
         Header: '',
         Cell: ({ row }) => (
           <Button variant="link" className="px-1" onClick={() => onViewMessageClick(tableData[row.index])}>
+            <span className="sr-only"> {row.index}</span>
             {intl.formatMessage(messages.buttonViewMessage)}
           </Button>
         ),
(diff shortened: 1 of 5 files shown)


cdeery (commented):
Have you gone back through the HTML to verify it is now correct?
Is there anything in here that should be checked in a unit test?

Tj-Tracy:
I'm not sure I understand the first question, the screenshots in the PR description show the HTML needed to meet each bullet point on the feedback.

I think the only thing we would need to add a unit test for is the extra `<span>` for screen readers, which I will add.

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
```diff
@@ -108,6 +108,7 @@ function BulkEmailContentHistory({ intl }) {
         Header: '',
         Cell: ({ row }) => (
           <Button variant="link" className="px-1" onClick={() => onViewMessageClick(tableData[row.index])}>
+            <span className="sr-only"> {row.index}</span>
```
wittjeff, line 111: Can you put this after the next line (so 'view message x' instead of 'x view message')? Actually from your screen capture it looks like maybe the intended space isn't making it in there before the X, so maybe <span className="sr-only">&nbsp{row.index}</span> ?

wittjeff (changes requested):

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailContentHistory.jsx
Tj-Tracy: looks like that fixed the markup 👍 :

image

wittjeff (approved):

PR #117, opened 2022-12-20
build: major version upgrade of paragon

Mashal-m:
Major version upgrade of `@edx/paragon`.


MaxFrank13 (approved):
Thank you for doing this work!

PR #125, opened 2023-05-19
feat: Support runtime configuration

arbrandes:
frontend-platform supports runtime configuration since 2.5.0 (see the PR that introduced it[1], but it requires MFE cooperation.  This implements just that: by avoiding making configuration values constant, it should now be possible to change them after initialization.

Only a single change related to the `LMS_BASE_URL` setting was required.

[1] openedx/frontend-platform#335

### DIFF
--- a/src/components/page-container/data/api.js
+++ b/src/components/page-container/data/api.js
@@ -1,10 +1,10 @@
 import { camelCaseObject, getConfig } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 
-const courseHomeBaseUrl = `${getConfig().LMS_BASE_URL}/api/course_home/v1/course_metadata`;
+export const getCourseHomeBaseUrl = () => `${getConfig().LMS_BASE_URL}/api/course_home/v1/course_metadata`;
 
 export async function getCourseHomeCourseMetadata(courseId) {
-  const courseHomeMetadataUrl = `${courseHomeBaseUrl}/${courseId}`;
+  const courseHomeMetadataUrl = `${getCourseHomeBaseUrl()}/${courseId}`;
   const { data } = await getAuthenticatedHttpClient().get(courseHomeMetadataUrl);
   return camelCaseObject(data);
 }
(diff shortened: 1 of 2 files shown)


arbrandes:
@ghassanmas, you might want to take a look at this one too.

arbrandes:
And @dcoa. :)

ghassanmas (commented):
I have tested the changes and it works, using the following tutor-plugins: 
`touch "$(tutor plugins printroot)"/communications.py`

```python
from tutor import hooks
from tutormfe.plugin import MFE_APPS

@MFE_APPS.add()
def _add_my_mfe(mfes):
    mfes["communications"] = {
        "repository": "https://github.com/arbrandes/frontend-app-communications",
        "port": 1984,
        "version": "runtime-config", # optional, will default to the Open edX current tag.
    }
    return mfes




hooks.Filters.ENV_PATCHES.add_item(
      (
          "openedx-lms-production-settings",
          """
FEATURES['ENABLE_NEW_BULK_EMAIL_EXPERIENCE'] = True
COMMUNICATIONS_MICROFRONTEND_URL = "{% if ENABLE_HTTPS %}https://{% else %}http://{% endif %}{{ MFE_HOST }}/communications"
MFE_CONFIG["SCHEDULE_EMAIL_SECTION"] = True
"""
      ),priority=100
  )



```
Enablding and rebuidling the image, and then following this [doc ](https://docs.openedx.org/en/latest/community/release_notes/bulk_email.html)to enable Bulk email. 

- [x] Email was sent successfully 

But found different bug, the return to course button didn't incldue coure-id as such the link pointed to 404 page: 

image

The `a` tag is pointing to `LMS_BAS/courses/courses/instructor#view-course-info`
Where actually it should point to `LMS_BAS/courses/<course-id>/instructor#view-course-info`

This bug might or might not be related to this change, if it's not then I think it looks good to me 👍🏾

ghassanmas:
I don't think the bug is related, I have opened an issue #126

dcoa (approved):

arbrandes:
Thanks @ghassanmas and @dcoa for the reviews!  Now just waiting for a maintainer/core contributor to approve.

PR #127, opened 2023-05-21
fix: getting course-id when public path is set, closes #126

### LINKED ISSUE #126: Course-id is incorrect when public path is not /
In the navigation compoenet which suppose to link back to instructor tab in the lms 

https://github.com/openedx/frontend-app-communications/blob/0c34d86610a216d67d145322ee0fb0b73c9cb643/src/components/navigation-tabs/BackToInstructor.jsx#L13 

The probelm is; `window.location.pathname.split('/')[2]` woud return course-id if public path is `/` the case _The typical case of edx_, while it would return `courses` if public path, is something, the case of tutor.   

## Suggest solution: 
- Use state to get course-id instead of link.  
- Figure out if public path is used, if it's used then get it course by getting the third index instead of the second.

This bug was found while checking this #125. e

ghassanmas:
This fixes #126

  This change, change the way course-id is used, is navigation
  componenet, before it resolved by guessing course-id index
  in the url, which would not be true if the public path is
  set. Since public path would shift the index of course-id
  in the url when set.

  Instead the course-id is resolved through react-router just
  like the container componenet, using the `useParams` hook.
  
  ## Testing:
  
  If public path is not set, it shouldn't affect the behaviour, however if pubic path is set to anything, the back button would not work unless checking to this change

### DIFF
--- a/src/components/navigation-tabs/BackToInstructor.jsx
+++ b/src/components/navigation-tabs/BackToInstructor.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { useParams } from 'react-router-dom';
 
 import { getConfig } from '@edx/frontend-platform';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
@@ -6,11 +7,13 @@ import { Button, Icon } from '@edx/paragon';
 import { ArrowBack } from '@edx/paragon/icons';
 
 export default function BackToInstructor() {
+  const { courseId } = useParams();
+
   return (
     <Button
       variant="tertiary"
       className="mb-4.5 ml-n4.5 text-primary-500"
-      href={`${getConfig().LMS_BASE_URL}/courses/${window.location.pathname.split('/')[2]}/instructor#view-course-info`}
+      href={`${getConfig().LMS_BASE_URL}/courses/${courseId}/instructor#view-course-info`}
     >
       <Icon
         src={ArrowBack}


openedx-webhooks:
Thanks for the pull request, @ghassanmas!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

Please let us know once your PR is ready for our review and all tests are green.

arbrandes (changes requested):
@ghassanmas, thanks for this!  It looks like this fixes it, but do you mind adding a couple of unit tests, one with the subpath set and one without?  This will make sure future changes don't break this for either use case.

arslanashraf7 (commented):
@ghassanmas While I was looking at a similar issue and I also noticed another place that would show a similar issue. 

There is a card at the bottom of the bulk email that provides a `Course Info` link and that also misses the course ID. Attaching the screenshot below.

While you're working on this, I thought I'd mention it here.

I guess [this is the place](https://github.com/openedx/frontend-app-communications/blob/0c34d86610a216d67d145322ee0fb0b73c9cb643/src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailPendingTasksAlert.jsx#LL16C77-L16C77) that might need a change as well. 

image

ghassanmas:
@arbrandes check the latest commit and let me know, if it looks good use for the other componenet.

src/components/page-container/test/PageContainer.test.jsx
```diff
@@ -64,4 +65,38 @@ describe('PageContainer', () => {
...
+      render(
+        <PageContainer>
+          <BackToInstructor />
+        </PageContainer>,
+      );
+      const linkEl = await screen.findByText('Back to Instructor Dashboard');
+      expect(linkEl.href).toEqual('http://localhost:18000/courses/test-course-id/instructor#view-course-info');
+    });
+  });
+  test('Page container with BackToInstructor compo with PUBLIC_PATH', async () => {
```
arbrandes, line 86: Instead of repeating tests that are similar except for a few variables (in this case, `PUBLIC_PATH`), I suggest you use `test.each()`.  Like [this instance in frontend-app-discussions](https://github.com/openedx/frontend-app-discussions/blob/master/src/discussions/discussions-home/BlackoutInformationBanner.test.jsx#L58-L63), for example.

Otherwise, this looks like a good place to include it!

src/components/navigation-tabs/BackToInstructor.jsx
```diff
@@ -1,16 +1,19 @@
...
 
 export default function BackToInstructor() {
+  const { courseId } = useParams();
+
   return (
     <Button
       variant="tertiary"
       className="mb-4.5 ml-n4.5 text-primary-500"
-      href={`${getConfig().LMS_BASE_URL}/courses/${window.location.pathname.split('/')[2]}/instructor#view-course-info`}
+      href={`${getConfig().LMS_BASE_URL}/courses/${courseId}/instructor#view-course-info`}
```
arbrandes, line 16: `courseId` should be passed in as a prop to this component.

It shouldn't be hard to do, as the only place that uses component has the variable [in scope](https://github.com/openedx/frontend-app-communications/blob/master/src/components/bulk-email-tool/BulkEmailTool.jsx#L16).

arbrandes (changes requested):
I think we need to re-architect what we're testing for, and where.  Apologies, it turns out that that is not a very good place to test, actually.

The good news it that I found a test that does something similar to what I was thinking.  See [this one in frontend-app-ora-grading](https://github.com/openedx/frontend-app-ora-grading/blob/master/src/data/constants/app.test.js#L15-L23).

Here, since what's affected by this currently is `BackToInstructor`, I think the test should be specific to this component, in a new `BackToInstructor.test.jsx`.  You can then try to mock `window.location.pathname` to be something that contains a PUBLIC_PATH and something that doesn't, and check that both work.

src/components/page-container/test/PageContainer.test.jsx
```diff
@@ -64,4 +65,38 @@ describe('PageContainer', () => {
...
+    await act(async () => {
+      const cohorts = { cohorts: [Factory.build('cohort'), Factory.build('cohort')] };
+      const courseMetadata = Factory.build('courseMetadata', {}, { host: 'http://localhost:18000/communication-path' });
```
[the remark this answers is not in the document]
ghassanmas: >  As I actually try to understand the test more deeply, it occurs to me to ask: how does courseMetadata.host affect the path?  

So my understanding is that when factory builds courseMetedata object,  it will pass the `host` to the `tab obecjt` hence here  https://github.com/openedx/frontend-app-communications/blob/4493eb71fa4fdaef3f5c541cff2d9f95d91ec14c/src/components/page-container/data/__factories__/courseMetadata.factory.js#L19-L29 

Of which tab object would create the url using  `course-id`,  `path` and `host`. hence 

https://github.com/openedx/frontend-app-communications/blob/4493eb71fa4fdaef3f5c541cff2d9f95d91ec14c/src/components/page-container/data/__factories__/tab.factory.js#L12 


> I think the right way to test this is going to be by mocking getConfig. More on that on a separate comment 

Hum... I think probably the right way to test it, to involve react-router, or at least override location vlaue from `useLocation` hook, and assuming devs won't use the `window` object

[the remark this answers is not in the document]
ghassanmas: I will look into, I don't how I missed your upper comment which is more relevant. 

> You can then try to mock window.location.pathname to be something that contains a PUBLIC_PATH and something that doesn't, and check that both work 

My understanding, is that if the future a dev by mistake replaced course-id that is taken from params, by window.location, the test should fail, right?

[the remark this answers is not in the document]
ghassanmas: @arbrandes you can check the last commit https://github.com/openedx/frontend-app-communications/pull/127/commits/e251025a70c6644ba235bfecc992d62242abcd6f  in the second test, I have changed window.locatoin.pathname value and test for both, course url and window.locaiotn.pathname.

src/components/navigation-tabs/BackToInstructor.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+    const linkEl = await screen.findByText('Back to Instructor Dashboard');
+    expect(linkEl.href).toEqual('http://localhost:18000/courses/test-course-id/instructor#view-course-info');
+  });
+
+  Object.defineProperty(window, 'location', {
+    get() {
+      return { pathname: '/communications/courses/test-course-id/bulk-email' };
+    },
+  });
```
arbrandes, line 31: This should probably be inside the `test()` below, no?

src/components/page-container/test/PageContainer.test.jsx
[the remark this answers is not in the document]
arbrandes: You can remove the new tests from `PageContainer`, now.

arbrandes (changes requested):
Made another pass, @ghassanmas.  Thanks for bearing with me!

arbrandes:
Oh, and can you please squash your commits so that the linter passes?

ghassanmas:
@arbrandes I think I have addressed them and just test the both urls after building the mfe, and it behave as expected.

src/components/navigation-tabs/BackToInstructor.jsx
```diff
@@ -1,16 +1,19 @@
 import React from 'react';
+import { useParams } from 'react-router-dom';
 
 import { getConfig } from '@edx/frontend-platform';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import { Button, Icon } from '@edx/paragon';
 import { ArrowBack } from '@edx/paragon/icons';
 
 export default function BackToInstructor() {
+  const { courseId } = useParams();
```
arbrandes, line 10: This should come in as a prop, so that `useParams()` is used in as few places as possible.

ghassanmas: Yeah, I don't have strong opinon on either, I though useParams, is simlair to global context. Is it better for optimization

src/components/bulk-email-tool/bulk-email-task-manager/BulkEmailPendingTasksAlert.jsx
```diff
@@ -1,19 +1,22 @@
 import React from 'react';
+import { useParams } from 'react-router-dom';
 
 import { getConfig } from '@edx/frontend-platform';
 import { Hyperlink, Alert } from '@edx/paragon';
 import { WarningFilled } from '@edx/paragon/icons';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
 
 export default function BulkEmailPendingTasksAlert() {
+  const { courseId } = useParams();
```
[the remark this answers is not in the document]
ghassanmas: That would be two level nested though

src/components/navigation-tabs/BackToInstructor.jsx
arbrandes: In this case, it means we'll have a single place in the code that defines courseId.  It makes it easier to debug and test.

Plus, this is what props are for.  :shrug:

arbrandes:
@ghassanmas, looks good, thanks!  Mind creating a backport to Palm?

ghassanmas:
@arbrandes just did #130  not there is also the  (#129) node v18 backport which we might need to merge first, given my changes here applied on top of Node 18 update thus tjhe backport if failiing node 16 tests, once #129 is merged I can rebease it.

PR #184, opened 2023-12-13
 feat: slot component for dynamic plugins

johnvente:
### Description

Following this [POC pluggable approach](https://github.com/openedx/frontend-app-course-authoring/pull/638) to importing plugins, we've created a reusable component that will dynamically import a module wherever it is called. The main idea is to be able to change the plugins with this mechanism, which means that instead of modifying the micro-frontend beforehand, we could change the plugin in isolation.

slot-compoentn

---
Using pluggable component with `BulkEmailForm` component

image

---
The alert is a plugin as well
image

#### How does it work?

Any component can be wrapped with the pluggable component. For instance:

```jsx
<PluggableComponent id="input-form" as="communications-app-card">
    <h1>Any component as default</h1>
</PluggableComponent>
```

The prop `as` indicates where is the plugin allocated. In this case, the component will be pointing to this route: `node_modules/@openedx-plugins/communications-app-card`.

If the plugin was installed as a dependency, it will render the component that you have pointing to that one.

If the plugin isn't installed, it will return its children. In this case, it's this:
```jsx
<h1>Any component as default</h1>
```

### Multiple plugins

We've added these props to `PluggableComponent`:

- `plugins`: This prop allows us to add multiple plugins, turning `PluggableComponent` into a container where its children will be rendered. Here's an example:

```jsx
import PluggableComponent from 'path-to/PluggableComponent';

const myPluginsList = [
  { id: 'plugin1', name: 'communications-app-test-component' },
  { id: 'plugin2', name: 'communications-app-test-component-2' }
];

const MultiplePlugins = () => (
  <PluggableComponent plugins={myPluginsList} />
);
```

If the name of the plugin, for example, is @openedx-plugins/communications-app-test-component, you should provide the name without @openedx-plugins, so it would be "communications-app-test-component".

- `pluginsPrefix`:  This prop allows us to add multiple plugins by specifying a prefix. PluggableComponent will find all plugins with that prefix and render them. For example, suppose you have three plugins:
    - @openedx-plugins/example-app-plugins-first-one
    - @openedx-plugins/example-app-plugins-second-one
    - @openedx-plugins/example-app-plugins-third-one
    
You only need to specify the prefix as a prop of PluggableComponent. In this case, "example-app-plugins" will render all plugins installed with that prefix. Here's an example:

```jsx
 import PluggableComponent from 'path-to/PluggableComponent';

  const MultiplePlugins = () => (
    <PluggableComponent pluginsPrefix="example-app-plugins"  />
  );

```
- `containerPluginsProps`: This prop allows us to specify props for the container of the plugins, such as a className, styles, or other props. Here's an example:

```jsx
 import PluggableComponent from 'path-to/PluggableComponent';

  const MultiplePlugins = () => (
    <PluggableComponent 
       pluginsPrefix="example-app-plugins" 
       containerPluginsProps={ className: "my_class", style: { backgroundColor: "red" } }  
    />
  );

```

#### Why are we doing this?

This means that the plugin doesn't have to be installed if I'm making a feature. It's something that we can improve in the future. That way, another developer can replace that component and make more changes. This approach allows us to wrap as many components as we can, making most of the features pluggable and adaptable to various use cases. The component can be rendered without children as well, which means it won't return anything if the plugin is not installed.

### How to test it

Check it out this branch and follow these steps:

```
 nvm use && npm install
```
Then you can run the mfe with `npm start`

If you want to create another plugin you can do it in the folder plugins/communications-app/ with the structure of the other
plugins

### Adding a Plugin That Lives Outside This Repository
You can do it by pointing to your npm package in your ``package.json``

```json
{
  "dependencies": {
     "@openedx-plugins/communications-app-body-email-form": "npm:@openedx/plugins-communications-app-body-email-form@^1.0.0"
  }
}
```
Or installing it like this

```bash
npm install --legacy-peer-deps "@openedx-plugins/communications-app-body-email-form@npm:@openedx/plugins-communications-app-body-email-form@^1.0.0"
```

To create and upload a plugin to npm you could follow this [guide](https://github.com/openedx/frontend-app-communications/blob/fc0b1bf2d18146fd626424009e8e92d10770ce87/docs/how_tos/create_an_external_plugin_ui_slot.rst)


openedx-webhooks:
Thanks for the pull request, @johnvente! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -267,6 +269,44 @@ function BulkEmailForm(props) {
         }}
       />
       <Form>
+        {/* this will be pluggable */}
+        <div className="border border-brand-400 p-3">
+          <h1>Pluggable components</h1>
+          <PluggableComponent
+            id="checkbox-form"
+            as="communications-app-check-box-form"
+            label="checkbox label - @openedx-plugins/communications-app-check-box-form"
```
johnvente, line 278: This will be for this plugin @openedx-plugins/communications-app-check-box-form

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -267,6 +269,44 @@ function BulkEmailForm(props) {
...
+            onChange={() => null}
+          >
+            <h1>Checkbox -default</h1>
+          </PluggableComponent>
+
+          <PluggableComponent
+            id="input-form"
+            as="communications-app-input-form"
```
johnvente, line 286: This will be for this plugin **@openedx-plugins/communications-app-input-form**

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -267,6 +269,44 @@ function BulkEmailForm(props) {
...
+            isValid
+            label="Hello"
+            feedbackText="This is okay"
+            controlId="test"
+          >
+            <h1>Input -default</h1>
+          </PluggableComponent>
+          {/* this will return default child if the plugin has not been installed  */}
+          <PluggableComponent id="input-form" as="communications-app-card">
```
johnvente, line 295: This points to this plugin: **@openedx-plugins/communications-app-card** but in this case that plugin doesn't exits so this will render the card above

src/components/PluggableComponent/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+    const loadPluginComponent = async () => {
+      try {
+        const hasModuleInstalled = await isPluginAvailable(as);
+
+        if (hasModuleInstalled) {
+          const PluginComponent = loadable(() => import(`@node_modules/@openedx-plugins/${as}`));
+
+          const component = children ? (
+            <PluginComponent key={id} {...pluggableComponentProps}>
+              {children}
```
bradenmacdonald, line 36: I like that the children are passed to the plugin, so it can optionally wrap the default contents 👍🏻

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -267,6 +269,44 @@ function BulkEmailForm(props) {
...
       />
       <Form>
+        {/* this will be pluggable */}
+        <div className="border border-brand-400 p-3">
+          <h1>Pluggable components</h1>
+          <PluggableComponent
+            id="checkbox-form"
+            as="communications-app-check-box-form"
+            label="checkbox label - @openedx-plugins/communications-app-check-box-form"
+            isChecked
```
bradenmacdonald, line 279: It seems a bit strange to me that these plugins are so specific. In this case a checkbox and an input box. What if I want to replace the check box with two checkboxes, or with an input box? It seems like that sort of flexibility is missing here. With props like `isChecked` being passed to the plugin, it seems like the checkbox plugin must render a checkbox. So isn't the customization too limited?

Also, I guess you're still working on this but it seems like `handleCheck` is missing.

johnvente: I agree. If you have any number of children, the props will be limited only to the plugin. However, we would need to pass props for each child in a prop called 'childrenProps' or something similar, which would be an object to specify its props for each one. But I don't like that the component's props will be too many if you have many components as children. I think each plugin should be responsible for a specific task. I was wondering if there is another way to handle that customization case

bradenmacdonald: I think in general the plugins should get a very simple prop that applies no matter how many widgets are in the slot, or they should get their data from React Context, not props.

Here is one idea for how you could implement something like that:

```ts
function BulkEmailForm(props) {

// Store the state of all the form fields. The fields are controlled by plugins.
// Each key in formFieldsState is a plugin ID, and each value is {value: any, isReady: boolean}
const [formFieldsState, updateFormFieldsState] = React.useState({});
const isReady = Object.keys(formFieldsState).length > 1 && Object.values(formFieldsState).every(field => field.isReady);
const pluginProps = {formFieldsState, updateFormFieldsState};

return <Form>
    <PluggableComponents id="bulk-email-form" {...pluginProps}>
        <PluggableComponent id="bulk-email-form-email-recipient" {...pluginProps}>
        <PluggableComponent id="bulk-email-form-email-subject" {...pluginProps}>
        <PluggableComponent id="bulk-email-form-email-body" {...pluginProps}>
    </PluggableComponents>
</Form>;
}
```

Then each plugin can manage its own state (checked or not) and data like this:
```
const pluginId = "new-checkbox-field";
const state = props.formFieldsState[pluginid] ?? {};
const updateState = (changes) => {
    props.updateFormFieldsState(oldState => ({...oldState, [pluginId]: {...oldState[pluginId], ...changes}));
};
if (!state.isReady) {
    updateState({isReady: true}); // The checkbox is not required, so allow users to submit the form whether or not this is checked.
}

return <Checkbox checked={state.value} onChange={(isChecked) => updateState({value: isChecked})} />
```

johnvente: Awesome! This is a really great idea, I will check this out. That way even each pluggableComponents can update the state of the parent and sometimes you will need to access to that data. Thanks a lot I'll be refactoring all the form that way

johnvente:
Hi guys! @bradenmacdonald @arbrandes now `BulkEmailForm` is full pluggable, I've changed it for `BuildEmailFormExtensible` I will be checking the tests

johnvente:
Hi there @arbrandes! This is ready to be merged now. I will update any change requests here

felipemontoya:
We used the version in this branch to build a few extensions in the Spanish project with great results. I'm very happy with the way this turned out.

@bradenmacdonald after your firsts comments were address, do you still have any remaining concerns?
@arbrandes I saw this in your list of pending reviews, just giving a friendly remainder to take a look when you can.

arbrandes:
I got back from break this week, will slot this in for a proper review early next one!  Exciting stuff!

bradenmacdonald:
@johnvente Does this implementation support multiple plugins in a single "slot"? For example on the sidebar of the studio Unit Page, or on the Learner Dashboard, we will almost certainly want to have some "slots" where several different plugins can appear together.

johnvente:
> @johnvente Does this implementation support multiple plugins in a single "slot"? For example on the sidebar of the studio Unit Page, or on the Learner Dashboard, we will almost certainly want to have some "slots" where several different plugins can appear together.

@bradenmacdonald In that case you could use one slot and inside of that one have multiple of them something like this

```
<PluggableComponent id="default-slot" as="plugin-app-container">
    <PluggableComponent id="sidebar-slot" as="plugin-app-sidebar">
                 <h1>Here it will load the sidebar </h1>
    </PluggableComponent>
    
     <PluggableComponent id="unit-slot" as="plugin-app-unit-course">
                 <h1>Here it will load the unit </h1>
    </PluggableComponent>
</PluggableComponent>
```

The children will be rendered unless the plugin-app-container is not installed other wise it will replace it's children. So yeah, it would be possible. Even inside each plugin you could have more plugins. Please feel free to ask any cause that you think can be implemented, I'll be glad to add them

bradenmacdonald:
@johnvente I mean something without having to define each plugin in the code.

Screenshot 2024-01-24 at 10 33 58 AM 2

Like here, could the green rectangle be a slot, which is at first empty. Then I install `plugin-course-promo` and the first widget appears, and I install `plugin-news` and the second widget appears, and then I install `plugin-braden-custom-progress` and the third plugin appears, and so on. Ideally I'd like to be able to add as many plugins as I want to this slot without having to change the code or create a meta-plugin. But listing them in the MFE configuration would be fine (`course_listing_sidebar_plugins = ["course-promo", "news", "braden-custom-progress"]`) or having them auto-detected when I `npm install` them would be fine. But I don't really want to have to modify the .jsx file of the dashboard nor create a "wrapper plugin" that contains all the other plugins.

arbrandes:
Just a quick inform: I'm currently attending Axim's in-person week, so won't have time to get back here until next week.

In the meantime, though, I find I agree with Braden's comments - in particular the last one.

johnvente:
> I mean something without having to define each plugin in the code

@bradenmacdonald You should add a wrapper and add there what you would like, remember that this implementation is only a component so that part doesn't have this one. We could add what you have mentioned before in another prop that if has defined some configuration then will render the configuration passed, what do you think?

bradenmacdonald:
@johnvente Could you give an example? I'm not sure I follow.

johnvente:
> @johnvente Could you give an example? I'm not sure I follow.

I mean currently you can create a wrapper like this:
`<PluggableComponent id="default-slot" as="plugin-app-wrapper" />`

Inside that component you could create as much plugins as you want and you won't have to rewrite inside the MFE, eventually that wrapper needs to be installed. Supuse that you have in that plugin something like this

```
  const courseSiteBarItems = ["course-promo", "news", "braden-custom-progress"] 
  const Wrapper = () => {
     return (
     <div>
         {courseSiteBarItems.map((courseItem) => <PluggableComponent id={uuid()} as={courseItem}  />)}     
     </div>)
  }
```

Then you could install them without updating another places just the `Wrapper`, but the wrapper needs to be in the MFE.
So in this part you can run after "npm install course-promo" or the other defined there. Now if you want to have plugins 

As you can see, you only will update the plugin-app-wrapper that contains another ones.  Now, if you want an special configuration so that part is that I mention that we could add that special configuration, some use cases that I couldn't cover and they are important to take account them. I think this part is a littler bit cleaner

johnvente:
Hi there guys! @bradenmacdonald @arbrandes I've some new props to support multiple plugins.

- `plugins`: This prop allows us to add multiple plugins, turning `PluggableComponent` into a container where its children will be rendered. Here's an example:

```jsx
import PluggableComponent from 'path-to/PluggableComponent';

const myPluginsList = [
  { id: 'plugin1', name: 'communications-app-test-component' },
  { id: 'plugin2', name: 'communications-app-test-component-2' }
];

const MultiplePlugins = () => (
  <PluggableComponent plugins={myPluginsList} />
);
```

If the name of the plugin, for example, is @openedx-plugins/communications-app-test-component, you should provide the name without @openedx-plugins, so it would be "communications-app-test-component".

- `pluginsPrefix`:  This prop allows us to add multiple plugins by specifying a prefix. PluggableComponent will find all plugins with that prefix and render them. For example, suppose you have three plugins:
    - @openedx-plugins/example-app-plugins-first-one
    - @openedx-plugins/example-app-plugins-second-one
    - @openedx-plugins/example-app-plugins-third-one
    
You only need to specify the prefix as a prop of PluggableComponent. In this case, "example-app-plugins" will render all plugins installed with that prefix. Here's an example:

```jsx
 import PluggableComponent from 'path-to/PluggableComponent';

  const MultiplePlugins = () => (
    <PluggableComponent pluginsPrefix="example-app-plugins"  />
  );

```
If you would like to have another plugin, you should install it, and it would be rendered there without having to update the code. For example, if you run
```
 npm install @openedx-plugins/example-app-plugins-another-one 
```
- `containerPluginsProps`: This prop allows us to specify props for the container of the plugins, such as a className, styles, or other props. Here's an example:

```jsx
 import PluggableComponent from 'path-to/PluggableComponent';

  const MultiplePlugins = () => (
    <PluggableComponent 
       pluginsPrefix="example-app-plugins" 
       containerPluginsProps={ className: "my_class", style: { backgroundColor: "red" } }  
    />
  );

```

For now `pluginsPrefix` I'm doing possible importing package.json with the relative path, that path should be of any MFE I think it would be great to import it like 'openedx/frontend-build' to be dynamically

brian-smith-tcril:
Hi @johnvente! Thank you so much for building this out! Pluggability is incredibly important for the platform, and it's wonderful to see the effort you've put into making it a reality!

I've started looking through the code on this. One thing that would be really helpful to aid in my understanding would be expanding upon the testing steps.

> ### How to test it
> 
> Check it out this branch and follow these steps:
> 
> ```
> nvm use && npm install
> ```
> Then you can run the mfe with `npm start`
> 
> If you want to create another plugin you can do it in the folder plugins/communications-app/ with the structure of the other
plugins

Adding instructions for installing/using a plugin that lives outside this repo would be incredibly helpful to clarify how this all fits together. Even better would be adding a how-to to [the docs directory](https://github.com/openedx/frontend-app-communications/tree/master/docs/how_tos)!

johnvente:
> Hi @johnvente! Thank you so much for building this out! Pluggability is incredibly important for the platform, and it's wonderful to see the effort you've put into making it a reality!
> 
> I've started looking through the code on this. One thing that would be really helpful to aid in my understanding would be expanding upon the testing steps.
> 
> > ### How to test it
> > Check it out this branch and follow these steps:
> > ```
> > nvm use && npm install
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then you can run the mfe with `npm start`
> > If you want to create another plugin you can do it in the folder plugins/communications-app/ with the structure of the other
> > plugins
> 
> Adding instructions for installing/using a plugin that lives outside this repo would be incredibly helpful to clarify how this all fits together. Even better would be adding a how-to to [the docs directory](https://github.com/openedx/frontend-app-communications/tree/master/docs/how_tos)!

Hi there @brian-smith-tcril! Thanks for taking a look over here. I agree I will be adding instructions to install a new plugin outside this repo

johnvente:
Hi @brian-smith-tcril! I've added a small section about how to install a plugin that lives outsite the MFE also a doc for creating and uploading one to npm

brian-smith-tcril:
@johnvente thank you so much for adding those docs! It seems to me they describe the process of *creating* a plugin, but it's still not clear how I would go about *consuming* an external plugin. Would I need to edit `BuildEmailFormExtensible` and add a new `PluggableComponent`, or is that use case out of scope for this PR?

If the scope of this PR is limited to replacing instead of adding, then simply copying what you have in the "Adding a Plugin That Lives Outside This Repository" section of the PR description into a how-to should cover the consuming aspect. The `npm install` instructions you have in the PR description seems to answer the question of "how do I modify the bulk email form without forking `frontend-app-communications`?" quite well.

Thanks again for this!

johnvente:
Hi @brian-smith-tcril! I think the main idea of this PR is to have a `PluggableComponent` that can be installed without forking the main MFE. We have used it in `BuildEmailFormExtensible` as an example, and we should be updating the plugins instead of the MFE. So, I guess that the form has enough plugins. If you want to add another plugin, you can do it inside any of the plugins that are currently available. Alternatively, you could have all the form components inside one plugin and not install the others. If you are referring "consuming an external plugin" like a plugin that works like cdn or a plugin with plugin from npm different to these plugins that is not covered in this PR

src/components/bulk-email-tool/BulkEmailTool.jsx
```diff
@@ -7,10 +7,11 @@ import { Container } from '@openedx/paragon';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import BulkEmailTaskManager from './bulk-email-task-manager/BulkEmailTaskManager';
 import NavigationTabs from '../navigation-tabs/NavigationTabs';
-import BulkEmailForm from './bulk-email-form';
+import BuildEmailFormExtensible from './bulk-email-form/BuildEmailFormExtensible';
```
brian-smith-tcril, line 10: @johnvente could you explain why both `BulkEmailForm` *and* `BuildEmailFormExtensible` exist in this PR? In my local testing I was able to delete `BulkEmailForm` and the page still worked (which I expected because it's only being consumed here https://github.com/search?q=repo%3Aopenedx%2Ffrontend-app-communications%20bulkemailform&type=code). I'm not sure if I'm missing something or if `BulkEmailForm` is just unused code at this point.

johnvente: @brian-smith-tcril At this point `BulkEmailForm` is unused code should I remove it?

brian-smith-tcril: That would be perfect! Thank you!

arbrandes:
@johnvente, it looks like there are a couple of comments of Brian's to address.

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
arbrandes: Was this addressed, yet?  Just checking.

src/components/bulk-email-tool/BulkEmailTool.jsx
johnvente: @brian-smith-tcril Done!

johnvente:
> @johnvente, it looks like there are a couple of comments of Brian's to address.

Hi @arbrandes sorry for the delay, I've made a refactor here with paragon dependency, I tried to remove the`BulkEmailForm` as @brian-smith-tcril [mentioned](https://github.com/openedx/frontend-app-communications/pull/184#discussion_r1488524529) here but the coverage was affected any idea? 

Thanks a lot. For now I will keep that component

brian-smith-tcril:
@johnvente please don't let codecov prevent you from removing `BulkEmailForm`. If coverage goes down we can look into the details of why and decide how to handle it after.

johnvente:
> @johnvente please don't let codecov prevent you from removing `BulkEmailForm`. If coverage goes down we can look into the details of why and decide how to handle it after.

Thanks! That said I've removed the `BulkEmailForm` component and other plugins as  `CheckBoxForm` and  `InputForm`

brian-smith-tcril:
@johnvente I haven't had a chance to give this another full review yet. I'll let you know when I do.

johnvente:
Hi guys @arbrandes @brian-smith-tcril [question] the pluggablecomponent will be something reusable, it would be better to have it in `frontend-platorm` or as part of `frontend-plugin-framework`? What do you guys think?

arbrandes:
@johnvente, the current idea is to merge this strategy with whatever is in https://github.com/openedx/frontend-plugin-framework/, and have it live there.  I imagine there are already a lot of similarities.  We'll just need to make sure this use-case is also served by it.

It will likely require some changes to the implementation here, but we always knew this might happen.  I encourage you to start looking at what's in the FPF repo linked above and how it could be used here - or even changed to be usable here.

felipemontoya:
@arbrandes I understand the approach of eventually merging the Widgets+UISlots and the Iframes-Plugins approaches. As I understand it however, that would mean merging this PR after the final review by @brian-smith-tcril (assuming nothing big comes up).
We are committed to bring pluggabillity to all MFEs, so I assume that both this implementation and that of iframes will evolve getting closer to one another. Only then it makes sense to me to move the pluggability code to frontend-platform and refactor communications to use that version instead of the local one.
Are we seeing the same path forward?

arbrandes:
Regarding the changes in this PR: as promised, we can merge this as soon as it passes review, yes.

But when it comes time to generalize the code, it should probably go to improving frontend-plugin-framework (as opposed to frontend-platform).  It doesn't just do iframes, by the way: it already has an implementation of UISlots itself, and there's at least one PR already open (oh wait, [it just merged](https://github.com/openedx/frontend-plugin-framework/pull/41)!) to make it even closer to what we have here.  When the time comes, we'll just have to see how much needs to be changed there, how much here.

(There are other things to see, as well: FPF currently requires the use of [Javascript configuration files](https://github.com/openedx/frontend-platform/blob/master/docs/decisions/0007-javascript-file-configuration.rst), which makes it very flexible, but tutor-mfe doesn't support it.  I've just started thinking about how to do it, and would love any input.)

Anyway, yeah, I think we're in alignment.  Let me know if y'all wanna have a call to discuss it explicitly.

arbrandes:
I added some ideas on how to support JS configuration here: https://github.com/overhangio/tutor-mfe/issues/199.  We should take that discussion there.

plugins/communications-app/BodyForm/index.jsx
```diff
@@ -0,0 +1,38 @@
...
+  const intl = useIntl();
+  const formData = useSelector((state) => state.form);
+  const dispatch = useDispatch();
+  const { body, isFormSubmitted = false } = formData;
+
+  const handleChangeTextEditor = (value) => {
+    dispatch(formActions.updateForm({ body: value }));
+  };
+
+  const isBodyValid = body.length > 0;
```
brian-smith-tcril, line 20: In the previous implementation, the same validation code is used to show/hide error messages *and* prevent `POST`ing.

https://github.com/openedx/frontend-app-communications/blob/6dd835d63f106c6e1bf7a02de16d3047a5bee558/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx#L153-L165

https://github.com/openedx/frontend-app-communications/blob/6dd835d63f106c6e1bf7a02de16d3047a5bee558/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx#L297

In this PR the logic has been split up and will need to be updated in multiple places if it ever changes. Have you considered ways to be able to utilize the same validation code for displaying errors and for preventing `POST`?

plugins/communications-app/RecipientsForm/index.jsx
```diff
@@ -0,0 +1,187 @@
...
+import './styles.scss';
+
+const disableIsHasLearners = ['track', 'cohort'];
+const recipientsFormDescription = 'A selectable choice from a list of potential email recipients';
+
+const RecipientsForm = ({ cohorts: additionalCohorts, courseModes }) => {
+  const formData = useSelector((state) => state.form);
+  const dispatch = useDispatch();
+  const { isEditMode, emailRecipients, isFormSubmitted } = formData;
+  const hasCourseModes = courseModes.length > 1;
```
brian-smith-tcril, line 17: In the current implementation this has null checking

https://github.com/openedx/frontend-app-communications/blob/6dd835d63f106c6e1bf7a02de16d3047a5bee558/src/components/bulk-email-tool/bulk-email-form/bulk-email-recipient/BulkEmailRecipient.jsx#L23

Is there a specific reason that was removed?

plugins/communications-app/RecipientsForm/index.jsx
```diff
@@ -0,0 +1,187 @@
...
+};
+
+RecipientsForm.propTypes = {
+  cohorts: PropTypes.arrayOf(PropTypes.string),
+  courseModes: PropTypes.arrayOf(
+    PropTypes.shape({
+      slug: PropTypes.string.isRequired,
+      name: PropTypes.string.isRequired,
+    }),
+  ),
```
brian-smith-tcril, line 184: In the current implementation the `courseModes` array is required

https://github.com/openedx/frontend-app-communications/blob/6dd835d63f106c6e1bf7a02de16d3047a5bee558/src/components/bulk-email-tool/bulk-email-form/bulk-email-recipient/BulkEmailRecipient.jsx#L130-L135

Is there a reason it was switched over to defaulting to an empty array instead?

plugins/communications-app/TaskAlertModalForm/index.jsx
```diff
@@ -0,0 +1,163 @@
...
+      setTimeout(() => handleResetFormValues(), 3000);
+    } catch {
+      changeFormStatus('error');
+    }
+  };
+
+  const createEmailTask = async () => {
+    const isScheduleValid = isScheduled ? scheduleDate.length > 0 && scheduleTime.length > 0 : true;
+    const isFormValid = emailRecipients.length > 0 && subject.length > 0
+    && body.length > 0 && isScheduleValid;
```
brian-smith-tcril, line 90: Same note as the one I left on `BodyForm` about having validation code in multiple places. I'd prefer to not require anyone in the future that updates the validation logic to make changes in more than one place.

src/components/bulk-email-tool/bulk-email-form/BuildEmailFormExtensible/index.jsx
```diff
@@ -0,0 +1,122 @@
...
+  useToggle,
+} from '@openedx/paragon';
+import { BulkEmailContext } from '../../bulk-email-context';
+import useMobileResponsive from '../../../../utils/useMobileResponsive';
+import PluggableComponent from '../../../PluggableComponent';
+
+import { withContextProvider, useDispatch } from './context';
+import { actionCreators as formActions } from './context/reducer';
+
+const BuildEmailFormExtensible = ({ courseId, cohorts, courseModes }) => {
```
brian-smith-tcril, line 17: Question about naming here: since this is completely replacing the current `BulkEmailForm`, does it make sense to simply use that name instead of using a new name? If not, and `BulkEmailForm` doesn't feel like an accurate name anymore, then I feel the `Extensible` is probably not needed and the name could simply be `BuildEmailForm`

brian-smith-tcril (commented):
Overall there's a lot of great stuff happening in here! It's wonderful to see how much of the old form has been split out into components that are easier to understand than they were when combined into one file!

I left a few comments with questions, looking forward to hearing your thoughts!

plugins/communications-app/RecipientsForm/index.jsx
johnvente: The reason is because `courseModes` are optional as prop and by default has an empty array so is not necessary to make the null validation

plugins/communications-app/RecipientsForm/index.jsx
johnvente: `courseModes` sometimes was undefined even if the prop is required so I think is better to pass a default value for that prop and remove the required

package.json
```diff
@@ -60,15 +68,20 @@
     "react-router-dom": "6.15.0",
     "redux": "4.2.0",
     "regenerator-runtime": "0.13.11",
-    "tinymce": "5.10.7"
+    "tinymce": "5.10.7",
+    "use-deep-compare-effect": "^1.8.1"
```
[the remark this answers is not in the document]
johnvente: Yep, even with that warning I think that lib worth here. since the `PluggableComponent` props will be an object I tried with [JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) inside useEffect array dependencies but the component broke sometimes. Another approche was using [`lodash.isEqual`](https://lodash.com/docs/4.17.15#isEqual) to compare props object but did not improve a lot. So at this point this lib is more solid that making something similar here. 

If you think this can be changed please let me know

plugins/communications-app/BodyForm/index.jsx
johnvente: I was wondering about use  [yup](https://github.com/jquense/yup) which allow to have an scheme to validate the fields that we need and then show the errors in the please required but I do not implemented here since the validations are only for the length of the fileds

src/components/bulk-email-tool/bulk-email-form/BuildEmailFormExtensible/index.jsx
johnvente: That makes sense I changed the name for `BulkEmailForm` as was

johnvente:
I won't be able to review this in a week so I will check any changes 1 April

plugins/communications-app/BodyForm/index.jsx
brian-smith-tcril: I'd prefer to avoid adding another library dependency, I was thinking something more along the lines of having the existing validation logic live in a file that we can import from in multiple places.

johnvente: Having account that we are using multiple plugins where place do you recommend? I wouldn't like to do it in the MFE now, in another plugin?

johnvente:
Hi @brian-smith-tcril I was wondering if checking the comments about the validations of the form in one place as this [comment](https://github.com/openedx/frontend-app-communications/pull/184#discussion_r1530708310) would it be enough or do you have more concerns? I'd like to check that or/those at the same time to merge this

mphilbrick211:
> Hi @brian-smith-tcril I was wondering if checking the comments about the validations of the form in one place as this [comment](https://github.com/openedx/frontend-app-communications/pull/184#discussion_r1530708310) would it be enough or do you have more concerns? I'd like to check that or/those at the same time to merge this

Just checking in on this, @brian-smith-tcril :)

arbrandes:
@felipemontoya, correct me if I'm wrong, but if I remember correctly we came to the conclusion that it would be best to refactor this using frontend-plugin-framework before merging it.  Can you confirm that's the case?

felipemontoya:
After there was a clear winner in terms of the framework we will use for frontend extension we did talk and decided it was not a very good outcome to have one MFE (this) with a different architecture of extension while everyother MFE would likely end up adopting FPF.

I did not close the PR because we are deploying this to some instances in the Spanish project and I don't want the branch to disappear before we actually have migrated it to FPF.

To reduce noise from this PR, could you please change it to `draft` so we can take the time to figure out the refactor effort?

arbrandes:
Thanks for the confirmation!  Converted it to draft.

mphilbrick211:
Hi @felipemontoya! Checking to see if this can be closed?

felipemontoya:
@mphilbrick211 yes. I'll close it now that we are no longer deploying to those instances.

PR #205, opened 2024-07-03
feat: same titles for sent email groups

khudym:
### Description
- this PR proposes making the display of mail recipients the same for all places (checkbox labels, **Send email** alert modal, **Scheduled emails** table with **View Modal**,  **Email Task History**)

**Related Pull Requests**
PR to the open-release/redwood.master branch: https://github.com/openedx/frontend-app-communications/pull/207
PR to the open-release/quince.master branch: https://github.com/openedx/frontend-app-communications/pull/206

### Steps to reproduce
click on "Staff/Administrators" checkbox -> complete the form (subject, body) -> options:
- to see recipients titles in **"Caution"** modal, click on "Send email" button
- same if you schedule day and time and click on "Schedule email" button
(in this modal you can see the difference between recipients titles and selected checkbox label)

-> click "Continue" button -> wait for the "Scheduled emails" table to load and also compare titles in "Send to" column

### BEFORE
Example with selected "Staff/Administrators"

Знімок екрана 2023-10-24 о 18 00 15

Знімок екрана 2023-10-24 о 18 00 28

Screenshot 2024-07-03 at 23 57 31

Знімок екрана 2023-10-24 о 18 02 22
<br>

Example with selected "All Learners"
<br>
Знімок екрана 2023-10-24 о 18 33 10

Знімок екрана 2023-10-24 о 18 32 57

Screenshot 2024-07-04 at 00 28 05

Знімок екрана 2023-10-24 о 18 33 16

### AFTER
Screenshot 2024-07-04 at 19 16 46

Screenshot 2024-07-04 at 19 15 34

Screenshot 2024-07-04 at 19 19 17

Screenshot 2024-07-04 at 19 23 00

Screenshot 2024-07-04 at 19 22 31

### DIFF
--- a/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
+++ b/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
@@ -28,6 +28,7 @@ import {
 } from './data/actions';
 import { editScheduledEmailThunk, postBulkEmailThunk } from './data/thunks';
 import { getScheduledBulkEmailThunk } from '../bulk-email-task-manager/bulk-email-scheduled-emails-table/data/thunks';
+import { getDisplayText } from '../utils';
 
 import './bulkEmailForm.scss';
 
@@ -219,7 +220,7 @@ function BulkEmailForm(props) {
       <p>{intl.formatMessage(messages.bulkEmailTaskAlertRecipients, { subject: editor.emailSubject })}</p>
       <ul className="list-unstyled">
         {editor.emailRecipients.map((group) => (
-          <li key={group}>{group}</li>
+          <li key={group}>{getDisplayText(group, courseModes)}</li>
         ))}
       </ul>
       {!isScheduled && (
@@ -246,7 +247,7 @@ function BulkEmailForm(props) {
       <p>{intl.formatMessage(messages.bulkEmailTaskAlertEditingTo)}</p>
       <ul className="list-unstyled">
         {editor.emailRecipients.map((group) => (
-          <li key={group}>{group}</li>
+          <li key={group}>{getDisplayText(group, courseModes)}</li>
         ))}
       </ul>
       <p>{intl.formatMessage(messages.bulkEmailTaskAlertEditingWarning)}</p>
--- a/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
+++ b/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
@@ -28,7 +28,7 @@ import {
 } from './data/actions';
 import { editScheduledEmailThunk, postBulkEmailThunk } from './data/thunks';
 import { getScheduledBulkEmailThunk } from '../bulk-email-task-manager/bulk-email-scheduled-emails-table/data/thunks';
-import { getDisplayText } from '../utils';
+import { getDisplayTextFromRecipient } from '../utils';
 
 import './bulkEmailForm.scss';
 
@@ -220,7 +220,7 @@ function BulkEmailForm(props) {
       <p>{intl.formatMessage(messages.bulkEmailTaskAlertRecipients, { subject: editor.emailSubject })}</p>
       <ul className="list-unstyled">
         {editor.emailRecipients.map((group) => (
-          <li key={group}>{getDisplayText(group, courseModes)}</li>
+          <li key={group}>{getDisplayTextFromRecipient(group)}</li>
         ))}
       </ul>
       {!isScheduled && (
@@ -247,7 +247,7 @@ function BulkEmailForm(props) {
       <p>{intl.formatMessage(messages.bulkEmailTaskAlertEditingTo)}</p>
       <ul className="list-unstyled">
         {editor.emailRecipients.map((group) => (
-          <li key={group}>{getDisplayText(group, courseModes)}</li>
+          <li key={group}>{getDisplayTextFromRecipient(group)}</li>
         ))}
       </ul>
       <p>{intl.formatMessage(messages.bulkEmailTaskAlertEditingWarning)}</p>
--- a/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
+++ b/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
@@ -220,7 +220,7 @@ function BulkEmailForm(props) {
       <p>{intl.formatMessage(messages.bulkEmailTaskAlertRecipients, { subject: editor.emailSubject })}</p>
       <ul className="list-unstyled">
         {editor.emailRecipients.map((group) => (
-          <li key={group}>{getDisplayTextFromRecipient(group)}</li>
+          <li key={group}>{getDisplayTextFromRecipient(intl, group)}</li>
         ))}
       </ul>
       {!isScheduled && (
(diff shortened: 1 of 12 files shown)


openedx-webhooks:
Thanks for the pull request, @khudym! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.

src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx
```diff
@@ -219,7 +220,7 @@ function BulkEmailForm(props) {
       <p>{intl.formatMessage(messages.bulkEmailTaskAlertRecipients, { subject: editor.emailSubject })}</p>
       <ul className="list-unstyled">
         {editor.emailRecipients.map((group) => (
-          <li key={group}>{group}</li>
+          <li key={group}>{getDisplayTextFromRecipient(group)}</li>
```
bradenmacdonald, line 223: This is not going to be internationalized, if I'm understanding this correctly.

You need to either:
1. Make `getDisplayTextFromRecipient(group)` return a `<FormattedMessage />` component, or
2. Make `getDisplayTextFromRecipient(group)` return a `MessageDescriptor` and then pass that to `intl.formatMessage(...)` as seen three lines above.

src/components/bulk-email-tool/bulk-email-form/bulk-email-recipient/BulkEmailRecipient.jsx
```diff
@@ -41,7 +42,7 @@ export default function BulkEmailRecipient(props) {
         <Form.Checkbox key="myself" value="myself" className="mt-2.5 col col-lg-4 col-sm-6 col-12">
           <FormattedMessage
             id="bulk.email.form.recipients.myself"
-            defaultMessage="Myself"
+            defaultMessage={RECIPIENTS_DISPLAY_NAMES.myself}
```
bradenmacdonald, line 45: I'm not sure but I think setting `defaultMessage` to a constant like this is going to break the extraction of strings for localization. In addition, it looks confusing because it looks like it's maybe a dynamic defaultMessage. Instead, all parts of this message descriptor (the `id`, the `defaultMessage`, and the `description`) should be put into a `messages` file and then you can just reference it here as `<FormattedMessage {...messages.recipientMyself} />`

src/components/bulk-email-tool/utils.js
```diff
@@ -0,0 +1,14 @@
+export const RECIPIENTS_DISPLAY_NAMES = {
+  myself: 'Myself',
+  staff: 'Staff and instructors',
+  learners: 'All students',
```
bradenmacdonald, line 4: I believe we generally prefer the term "Learner" to "Student" so this seems like a step backwards ?

src/components/bulk-email-tool/bulk-email-form/test/BulkEmailForm.test.jsx
```diff
@@ -75,8 +76,8 @@ describe('bulk-email-form', () => {
       success: true,
     });
     render(renderBulkEmailForm());
-    fireEvent.click(screen.getByRole('checkbox', { name: 'Myself' }));
-    expect(screen.getByRole('checkbox', { name: 'Myself' })).toBeChecked();
+    fireEvent.click(screen.getByRole('checkbox', { name: RECIPIENTS_DISPLAY_NAMES.myself }));
+    expect(screen.getByRole('checkbox', { name: RECIPIENTS_DISPLAY_NAMES.myself })).toBeChecked();
```
bradenmacdonald, line 80: Nit: It's fine to hard-code strings in test files; it makes it easier to see when you've accidentally changed something, like accidentally setting the value of `RECIPIENTS_DISPLAY_NAMES.myself` to "Learners".

src/components/bulk-email-tool/bulk-email-task-manager/bulk-email-scheduled-emails-table/BulkEmailScheduledEmailsTable.jsx
```diff
@@ -44,8 +45,8 @@ function BulkEmailScheduledEmailsTable({ intl }) {
   const [currentTask, setCurrentTask] = useState({});
 
   useEffect(() => {
-    setTableData(flattenScheduledEmailsArray(scheduledEmailsTable.results));
-  }, [scheduledEmailsTable.results]);
+    setTableData(flattenScheduledEmailsArray(scheduledEmailsTable.results, courseModes));
```
bradenmacdonald, line 48: I don't understand the addition of `courseModes` to this table. It's not mentioned in the PR description nor seems to be used anywhere ?

src/components/bulk-email-tool/bulk-email-task-manager/bulk-email-scheduled-emails-table/BulkEmailScheduledEmailsTable.jsx
```diff
@@ -96,7 +97,7 @@ function BulkEmailScheduledEmailsTable({ intl }) {
       },
     } = row;
     const dateTime = new Date(taskDueUTC);
-    const emailRecipients = targets.replaceAll('-', ':').split(', ');
+    const emailRecipients = targets.replaceAll('-', ':').split(', ').map(getRecipientFromDisplayText);
```
bradenmacdonald, line 100: Reversing the "display text" like this is not a very robust solution. Can you just modify `flattenScheduledEmailsArray` so that it stores _both_ the `email.courseEmail.targets` IDs and the display text on `row.original`  ? Then you won't have to do this reversing.

bradenmacdonald (changes requested):
I'm not very familiar with this code, but I took a pass over this for you. In the future, I'd recommend waiting until the `master` PR has been reviewed and merged before you open the backport PRs.

mphilbrick211:
@khudym hi there! Just checking in to see if this is still in-progress?

src/components/bulk-email-tool/utils.js
```diff
@@ -0,0 +1,21 @@
...
+ * Retrieves the display text for a given recipient.
+ *
+ * @param {Object} intl - The internationalization object, provided by React Intl.
+ * @param {string} recipient - The recipient key used to look up the corresponding display name.
+ * @returns {string} - The formatted display name for the recipient,
+ * or the original recipient key if no display name is found.
+ */
+// eslint-disable-next-line import/prefer-default-export
+export const getDisplayTextFromRecipient = (intl, recipient) => (
+  intl.formatMessage(RECIPIENTS_DISPLAY_NAMES[recipient]) || recipient
```
bradenmacdonald, line 20: Did you test this? I'm thinking that `intl.formatMessage` is going to throw an error if passed `undefined`, so I would write this as

```suggestion
  const msg = RECIPIENTS_DISPLAY_NAMES[recipient];
  return msg ? intl.formatMessage(msg) : recipient; // Fall back to the recipient key if no display name is found.
```

bradenmacdonald (commented):
Thanks for those changes. This looks fine to me (besides one last suggestion), but I'm not familiar with this app and don't know how to test it. @mphilbrick211  Is someone else able to finish this review? I was just trying to help with some suggestions, not take on the review myself.

mphilbrick211:
@openedx/committers-frontend - is someone able to finish this review?

mphilbrick211:
Friendly ping on this @openedx/committers-frontend

arbrandes:
Putting this on my list to review/test.

